Issue #465, opened 2015-12-22
Add support for pre-gzipped static files


SkeLLLa:
Hi, @tshemsedinov.
It would be nice to have pre-gzipped static files support. For example, if there is a file in static directory with *.gz extension, then it should be added as gzipped assert into application file cache.

This would be useful for compressing quite large files or compression with zopfli which takes a lot of cpu time. Also e.g. if there are a lot of servers in cluster it will be better to compress static files once (for example in some CI tool like jenkins) and distribute it all over the servers instead of gzipping all on each server.

I believe this can be done in `impress.application.js`, maybe somewhere in `compress` function https://github.com/tshemsedinov/impress/blob/master/lib/impress.application.js#L598. But I'm not sure what is the best way to do this: whether add check for existing file with same name, but with `.gz` extension before calling `api.zlib.gzip` or it can be done earlier in separate function which will be called on application start or static cache initialization. Also if you want I can submit a pull request, but then I need to know which option is better from your point of view.

tshemsedinov:
I think we can do this putting gzipped file to ```/static```, i.e.: ```/static/js/file.js.gz``` and addressing them like this: ```<script src="/js/file.js"></script>```. To do this we need to change ```Client.prototype.static``` in this place: https://github.com/tshemsedinov/impress/blob/master/lib/impress.client.js#L827 so if ```/static/js/file.js``` not exists we will try to find ```/static/js/file.js.gz``` to send back and put to memory cache.

Another way to do that is to put uncompressed files to ```/client``` and use CI tools to build it into  ```/static``` so we may stop using ```/static``` for uncompressed, even more, for not precompiled and not minified files.

You can do this and send me PR if you have time or I can do this later.

SkeLLLa:
@tshemsedinov in my case uncompressed files are in client folder, then all of the files are built with gulp and copied to static folder (including both uncompressed, e.g. index.html and compressed - index.html.gz). In real world those files could be even in separate repository.

I think that maybe it will be better if we check `/static/js/file.js.gz` first and then `/static/js/file.js` as "fallback", so it will be compressed version first.

Also may be it will be good idea to implement such behavior for almost all static files except images, because they are gzipped very rarely and, e.g. using `zopflipng` images will be aleady compressed inside png container, so gzipping them one more time will be redundant.

I'll take a look at this in few days, and if I'll be succesfull with this one, I'll submit a PR.

tshemsedinov:
[...]

PR #466, opened 2015-12-23
Add static pre-gzipped files support

SkeLLLa:
Relates to #465 

Implemented support for pre-gzipped files. Now static method tries to find compressed file first (with `.gz` extension). If it fails it fallbacks to previous behavior, and if it finds gzipped file, then it serves file and adds to cache.

@tshemsedinov please review if all is ok. I've run some test and all seems good. 
Also maybe it will be better to make some function like `serveAndCache` for code block:
```
api.fs.readFile(filePath + suffix, function(err, data) {
        if (err) client.error(404);
        else {
          client.stats = stats;
          client.compressed = true;
          client.end(data);
          application.addToCache(relPath, stats, true, data)
        }
      });
```
Or may be there is one, but I didn't find it.

### DIFF
--- a/lib/impress.client.js
+++ b/lib/impress.client.js
@@ -815,40 +815,71 @@ Client.prototype.runScript = function(handler, fileName, callback) {
 //
 Client.prototype.static = function(callback) {
   var client = this,
-      application = client.application;
+      application = client.application,
+      suffix = '.gz';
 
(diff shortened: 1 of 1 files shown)


tshemsedinov:
I am optimizing and testing it now, wait 1-2 days plz before further changes @SkeLLLa

SkeLLLa:
@tshemsedinov, ok, no problem. Thanks.

PR #469, opened 2015-12-28
Fix bug in API serving #468

### LINKED ISSUE #468: Bug in API serving
Static refactored and works well but AJAX API, WS and SSE handlers fails

SkeLLLa:
If `relPath` is not from static, then it should call callback function `onNotServed` to be able to proceed to server code, instead of trying to create index directory for it as it was done before adding static gz support.

This should fix #468.


tshemsedinov:
Yes we have a bug here but we can't use regular expressions in Impress core because of performance reduction they bring. I have solved this in different way and will publish in a few days.

SkeLLLa:
@tshemsedinov I think simple indexOf instead of regexp will solve the problem. Or you can add a flag variable that will indicate that this function was called for index file.

PR #580, opened 2016-08-08
Update impress.cms.schema.js

DzyubSpirit:
Just typing mistake.


DzyubSpirit:
Also this shows that impress.cms.schema.js doesn't corresponds impress.cms.schema.sql because
sql file has this LanguageId index.
https://github.com/metarhia/Impress/blob/master/schemas/impress.cms.schema.sql#L115

DzyubSpirit:
or not?

PR #585, opened 2016-08-30
Added connection to mysql via socket

IgorKalinchuk:


lib/api.db.js
```diff
@@ -75,7 +75,7 @@ api.db.openApplicationDatabases = function(application, callback) {
 api.db.openDatabase = function(application, database, callback) {
   if (!application.databases) application.databases = {};
   database.application = application;
-  database.schema = database.url.substr(0, database.url.indexOf(':'));
+  database.schema = (database.url) ? database.url.substr(0, database.url.indexOf(':')) : 'mysql';
```
tshemsedinov: mysql by default, why not mongo or pg?

lib/api.db.mysql.js
```diff
@@ -28,7 +28,7 @@ if (api.mysql) {
   api.db.mysql.open = function(database, callback) {
     database.retryCounter++;
     var application = database.application,
-        connection = api.mysql.createConnection(database.url);
```
tshemsedinov: Everything should be connection in connection string url

IgorKalinchuk: When msql configured with closed TCP/IP. I needed to use connection via socet, so I can't use string url in this case.
socketPath: The path to a unix domain socket to connect to. When used host and port are ignored.

lib/api.db.js
IgorKalinchuk: Yes, but as I understand only in mysql, can't be used string url

PR #591, opened 2016-09-09
Add mysql pool connection, connection via socket

IgorKalinchuk:


aqrln:
@tshemsedinov what about this PR? I think it is obsolete anyway since we plan to drop support for all DBMS except GlobalStorage. Though it is a right thing to communicate with a database server via Unix domain socket instead of network socket if the application and database servers are on the same machine. It should be implemented in GlobalStorage for its DBMS backends, namely the MongoDB provider which is now present. The Unix socket for MongoDB is `/tmp/mongodb-27017.sock` by default.

@IgorKalinchuk can you describe the other changes besides connection via Unix socket as the intention is not very clear from the code? What problem did you solve? Maybe it is worth to implement it in GlobalStorage too so you could send a PR there.

IgorKalinchuk:
@aqrln 
Hi! other changes: I added the ability to open connections to pull mySql database. This is a standard driver functionality for node.js

tshemsedinov:
* @aqrln we will support direct connections to DBMS
* Impress security provider will support only GlobalStorage providers
* we need no GlobalStorage provider for MySQL but we need it for Impress
* after Impress refactoring we have multiple code changes, so @IgorKalinchuk can you please rebase your changes and squash them into single commit?

Issue #613, opened 2016-10-26
Fix executing /setup/*.js files


tshemsedinov:

tshemsedinov:
no bug found

PR #615, opened 2016-10-29
fix setup files

JWo1F:
issue #613


tshemsedinov:
no bug, no need to fix

Issue #684, opened 2017-02-02
Review `delete` usage


aqrln:
Actually, `delete obj.prop` is pretty useless in ES2015. If an object is an object, nothing should be deleted. If an object is a hash map, then a `Map` should be used.

PR #694, opened 2017-02-15
Use ES6 Map, issue #684

tshemsedinov:
Here are last places where we can use ES6 Map. All other places are already refactored or it isn't a good case for using Map.

@aqrln take a look at `impress.jstp.js`

### DIFF
--- a/lib/impress.sse.js
+++ b/lib/impress.sse.js
@@ -20,14 +20,12 @@ impress.sse.mixin = (application) => {
   };
 
   application.sse.heartbeat = api.timers.setInterval(() => {
-    let connectionId, connection;
     const connections = application.sse.connections;
-    for (connectionId in connections) {
-      connection = connections.get(connectionId);
+    connections.forEach((key, connection) => {
       if (connection.heartbeat && !connection.res.finished) {
         connection.res.write(SSE_HEARTBEAT);
       }
-    }
+    });
   }, SSE_HB_INTERVAL);
 
   impress.Client.prototype.sseConnect = function(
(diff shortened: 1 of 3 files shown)


lib/impress.sse.js
```diff
@@ -136,7 +136,7 @@ impress.sse.mixin = (application) => {
     const connections = application.sse.connections;
     let connectionId, res;
     for (connectionId in connections) {
```
aqrln: We shouldn't had used a for-in loop before these changes because of performance issues, but now this won't even work correctly.

lib/impress.sse.js
```diff
@@ -23,7 +23,7 @@ impress.sse.mixin = (application) => {
     let connectionId, connection;
     const connections = application.sse.connections;
     for (connectionId in connections) {
```
aqrln: The same issue as reported below.

aqrln:
@tshemsedinov `impress.jstp.js` LGTM, `impress.sse.js` should be fine too after loops are fixed.

lib/impress.sse.js
```diff
@@ -22,12 +22,11 @@ impress.sse.mixin = (application) => {
   application.sse.heartbeat = api.timers.setInterval(() => {
     let connectionId, connection;
     const connections = application.sse.connections;
-    for (connectionId in connections) {
-      connection = connections.get(connectionId);
+    connections.forEach((key, connection) => {
```
aqrln: The signature of `Map#forEach` is `(value, key, map)`, so the parameters of this callback must be swapped to `(connection, key)` which translates into just `(connection) => { ... }` since `key` isn't used here.

lib/impress.sse.js
```diff
@@ -22,12 +22,11 @@ impress.sse.mixin = (application) => {
   application.sse.heartbeat = api.timers.setInterval(() => {
     let connectionId, connection;
```
aqrln: This line should be removed now.

lib/impress.sse.js
```diff
@@ -134,11 +133,10 @@ impress.sse.mixin = (application) => {
   ) => {
     const buf = impress.sse.packet(eventName, data);
     const connections = application.sse.connections;
-    let connectionId, res;
-    for (connectionId in connections) {
-      res = connections.get(connectionId).res;
+    connections.forEach((key, connection) => {
```
aqrln: Same as above, s/`(key, connection)`/`(connection)`.

Issue #680, opened 2017-02-02
Proper error handling


aqrln:
In many cases errors are not logged or handled properly or even silently ignored.

tshemsedinov:
Done in v2.x

PR #695, opened 2017-02-17
Version 0.4.11

tshemsedinov:


lib/impress.application.js
```diff
@@ -311,6 +311,7 @@ impress.application.mixin = (application) => {
         api.fs.stat(filePath, (err, stats) => {
           if (err || !stats.isFile()) return cb();
           application.createScript(filePath, (err, exports) => {
+            if (err) return cb();
```
aqrln: It's better than nothing, but ideally an error should bubble to the callback.

aqrln: And from what I see, `callback` actually accepts an error as its first parameter.

lib/impress.application.js
```diff
@@ -370,6 +371,7 @@ impress.application.mixin = (application) => {
     if (fileExt !== '.js') return callback();
     if (application.config[sectionName]) return callback();
     application.createScript(configFile, (err, exports) => {
+      if (err) return callback();
```
aqrln: The same issue, I think.

lib/api.db.gs.js
```diff
@@ -24,10 +24,7 @@ if (api.gs) {
         config.connection = db.connection;
       }
       api.gs.open(config, (err) => {
-        if (err) {
-          application.log.error('Can\'t open database: ' + database.name);
-          return;
-        }
+        if (err) return;
```
aqrln: Is it intentional that errors are now passed silently?

belochub:
@aqrln, #680 seems to be closed, even though after your review on this PR it is clear that this issue is not completely resolved. Should it be reopened?

lib/api.db.mysql.js
```diff
@@ -140,10 +140,10 @@ if (api.mysql) {
               }
             }
             fieldDef = fieldType.type;
-            if ('size' in field) fieldDef += '(' + field.size + ')';
+            if (field.size) fieldDef += '(' + field.size + ')';
             fieldDef += ' ';
-            if (('signed' in field) && !field.signed) fieldDef += 'unsigned ';
-            if ((('nullable' in field) && !field.nullable) || field.master) {
+            if (field.signed && !field.signed) fieldDef += 'unsigned ';
```
belochub: That is a really strange condition, does it ever evaluate to true?
And even if it does, I think you should change it to be more clear about your intentions.

lib/api.db.mysql.js
```diff
@@ -140,10 +140,10 @@ if (api.mysql) {
...
+            if ((field.nullable && !field.nullable) || field.master) {
```
belochub: Ditto.

lib/api.db.mysql.js
```diff
@@ -190,16 +190,16 @@ if (api.mysql) {
...
                 altr + tableName + ' ADD CONSTRAINT fk' + tableName +
                 fieldName + ' FOREIGN KEY (' + fieldName + ') REFERENCES ' +
                 reference.dataset + ' (' + referenceKey + ') ON DELETE '
               );
-              if (!('deletion' in reference)) {
+              if (!reference.deletion) {
                 if (field.master) {
                   reference.deletion = 'cascade';
-                } else if (('nullable' in field) && !field.nullable) {
+                } else if (field.nullable && !field.nullable) {
```
belochub: Here as well.

lib/impress.log.js
```diff
@@ -234,20 +229,14 @@ impress.log.mixin = (application) => {
...
-        fn = fnN;
-      } else if (!serverLog && applicationLog) {
-        fn = fnA;
-      } else if (serverLog && !applicationLog) {
-        fn = fnI;
-      } else if (serverLog && applicationLog) {
-        fn = fnAI;
-      }
-      application.log[fileType] = fn;
+      application.log[fileType] = (
```
belochub: It became much harder to read now (it was not easy before as well), you might as well rewrite it like that:
```javascript
const logFunctions = [fnN, fnA, fnI, fnAI, fnN, fnN, fnI, fnI];
let index = application.isImpress << 2 |
            serverLog << 1 |
            applicationLog << 0;
application.log[fileType] = logFunctions[index];
```
It does not add clarity, just feels like a more neat way to write conditions looking like this one.
I don't really know if there is any better way to do this, so that it is clear what this condition does at all, because I still don't understand it.

lib/api.db.mysql.js
aqrln: The diff has the answer about the intention. Before the change the condition checked that the field exists and evaluates to true, so the equivalent change would be just `field.signed === false` (if it's guaranteed to be a boolean), or, if the exact data type is unknown, `field.signed !== undefined && !field.signed` or `field.hasOwnProperty('signed') && !field.signed`.

lib/impress.application.js
```diff
@@ -278,13 +278,17 @@ impress.application.mixin = (application) => {
     api.fs.readdir(absPath, (err, files) => {
       if (err) {
         application.log.error(impress.CANT_READ_DIR + absPath);
-        return callback(err);
+        return callback();
```
aqrln: I'm not sure about this change, though it is the matter of https://github.com/metarhia/Impress/issues/680 anyway.

lib/impress.log.js
tshemsedinov: I think we need to find a way to rewrite it into clear code. See #701

lib/impress.application.js
tshemsedinov: No need to rise this logged error into callback.

lib/api.db.mysql.js
tshemsedinov: I added default value for this field, so we can use just `!field.nullable` here

lib/api.db.mysql.js
tshemsedinov: Default value added, so we can use just `!field.signed` and `!field.nullable`

lib/api.db.mysql.js
tshemsedinov: Default value arranges everything here too

Issue #697, opened 2017-02-17
Outdated usage of metarhia-jstp module in examples


belochub:
Examples of websocket connection to server do not work with latest metarhia-jstp versions due to changes in the API.

PR #698, opened 2017-02-17
Use symlinks to JSTP browser libs

aqrln:
Instead of copying browser JSTP libraries from the JSTP distribution on `npm install`, use symlinks to them so that the client-side part of `metarhia-jstp` is always up-to-date when running an example application and always has the same version as Impress uses at server side.

This kinda refs https://github.com/metarhia/Impress/issues/697. Though the described bug is caused by another issue, a user could theoretically run into similar behavior because of this issue too (though the symptoms would be rather opposite: new interface in the example application and old interface in the included JavaScript library).

### DIFF
--- a/install.js
+++ b/install.js
@@ -44,29 +44,37 @@ function installCLI() {
   });
 }
 
-// Copy the browser version of JSTP
+// Symlink the browser version of JSTP
 //
-['jstp.min.js', 'jstp.min.js.map'].forEach((file) => {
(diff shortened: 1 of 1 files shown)


aqrln:
I suppose we haven't introduced code review for each commit in Impress yet, so I rather opened this PR to look at CI results safely because of the nature of the patch. But if no one has objections to it, we may start a PR-based workflow with this commit :)

aqrln:
Great. Landed in 6a23a6346ea481865f97e63c8e74d0ec99f33836.

PR #700, opened 2017-02-18
Cache opt

tshemsedinov:


lib/api.common.js
```diff
@@ -594,18 +594,17 @@ api.common.sortCompareByName = (
   return 0;
 };
 
-api.common.clearCacheStartingWith = (
+api.common.clearCacheStartsWith = (
```
aqrln: I liked the previous name more. @belochub, what's your opinion?

aqrln: OTOH, the previous name may seem to be somehow connected to time, not name prefixes.

aqrln: Maybe it is even more appropriate to indicate somehow that _names_ are starting with something.

lib/impress.cache.js
```diff
@@ -133,66 +132,58 @@ impress.cache.mixin = (application) => {
...
-            key !== filePath &&
-            key.startsWith(filePath) &&
-            !files[key]
-          ) {
-            files[key] = stats;
+        // Directory changed
+        let key, starts, includes;
+        for (key of cache.files.keys()) {
+          starts = key.startsWith(filePath);
+          includes = files.includes(key);
```
aqrln: Maybe `files` can be a `Set`?

aqrln:
I'll review it more thoroughly later.

lib/api.common.js
aqrln: Ugh, never mind. For some reason I only reviewed the first commit, not the whole PR.

lib/impress.cache.js
tshemsedinov: Sure

lib/api.common.js
tshemsedinov: I removed this function at all, but Maps coming from factory `api.common.cache()` have `add`, `del` and `clr` methods

aqrln: Yeah, I got it, that's what my last comment is about :)

lib/api.common.js
```diff
@@ -657,39 +643,76 @@ api.common.eventEmitter = () => {
 
 api.common.cache = (
   // Extend Map interface total allocated size: map.allocated
+  options
 ) => {
   const cache = new Map();
+  const calcSize = (options && options.calcSize) || false;
```
aqrln: I'd use `0` instead of `false` here so that the data types are consistent.

lib/api.common.js
```diff
@@ -657,39 +643,76 @@ api.common.eventEmitter = () => {
...
+
+};
+
+api.common.cache.withSize = (
+  cache // mixin to this Map
+) => {
+
   cache.allocated = 0;
 
+  const dataSize = data => (data && data.lenght ? data.length : 0);
```
aqrln: `length`, not `lenght`

lib/api.common.js
aqrln: Never mind, I'd thought I was a number before I read the code further. Maybe it's indicating that the variable name isn't that good though.

lib/api.common.js
aqrln: This typo will cause `dataSize` to be always `0`.

lib/api.common.js
```diff
@@ -657,39 +643,76 @@ api.common.eventEmitter = () => {
...
     if (cache.has(key)) {
       const val = cache.get(key);
-      cache.allocated -= api.common.cacheSize(val);
+      cache.allocated -= dataSize(val);
     }
   };
 
-  cache.clearStartingWith = (
-    startsWith // string to compare with key start
+  cache.clr = (
```
aqrln: I believe it is duplicate code.

lib/api.common.js
```diff
@@ -657,39 +643,76 @@ api.common.eventEmitter = () => {
...
+    prefix, // string to compare with key start
+    fn // function(key, val) to be called on each key (optional)
   ) => {
-    let key, val;
-    for ([key, val] of cache) {
-      if (key.startsWith(startsWith)) {
-        cache.allocated -= api.common.cacheSize(val);
+    cache.forEach((val, key) => {
```
aqrln: Despite that it has inconsistent loop style :)

lib/api.common.js
tshemsedinov: Not duplicate, forEach and for..of

lib/api.common.js
tshemsedinov: What are you talking about?

lib/api.common.js
aqrln: Ah, I didn't notice it mutates `cache.allocated`. Never mind then :)

lib/api.common.js
aqrln: I thought the only difference was `for-of` vs `forEach`, but it actually needs `val`, so yeah, everything's okay here.

PR #713, opened 2017-03-28
Code style and error/callback handling optimizations

tshemsedinov:


lib/impress.files.js
```diff
@@ -31,77 +31,80 @@ Client.prototype.download = function(
...
 
   let fileCount = 0;
   let doneCount = 0;
 
-  const done = (data) => {
+  const cb = (data) => {
     doneCount++;
-    if (onFile) onFile(data);
-    if (fileCount === doneCount && onDone) onDone(doneCount);
+    if (each) each(data);
```
aqrln: Not related to this PR, but wouldn't it be reasonable to allow `each` to be an asynchronous function (and, therefore, run `done` only when it has invoked its callback)? It could also return a boolean value to indicate whether the file should be uploaded or rejected. Example use case: allowing files to upload only when there's a certain configurable amount of storage free.

aqrln: Well, I guess exactly that use case isn't much viable since the file is already on the disk, but I think you got the idea :)

lib/impress.files.js
```diff
@@ -31,77 +31,80 @@ Client.prototype.download = function(
...
-  onDone // optional callback function(doneCount)
+  done // optional callback function(doneCount)
 ) {
   const client = this;
 
   let fileCount = 0;
   let doneCount = 0;
 
-  const done = (data) => {
+  const cb = (data) => {
```
aqrln: It would be nice if `client.uploadFile()` passed an error as the first argument of its callback, not just log it. We could call `done` passing the count of successful uploads only, or even expose information about both successful and failed uploads.

lib/impress.files.js
```diff
@@ -31,77 +31,80 @@ Client.prototype.download = function(
...
+      client.uploadFile(file, cb);
     }
   }
 };
 
 function saveUploadedFile(
   // Save uploaded file
   data, // { compressionFlag, storagePath, storageSize }
-  onFile // function
+  each // function
```
aqrln: I guess `each` isn't a very appropriate name since there's only one file, not a collection of them.

lib/impress.files.js
```diff
@@ -31,77 +31,80 @@ Client.prototype.download = function(
...
 ) {
-  if (data.compressionFlag === 'N') return onFile(data);
+  if (data.compressionFlag === 'N') return each(data);
   api.fs.unlink(data.storagePath, () => {
     api.fs.rename(data.storagePath + '.tmp', data.storagePath, () => {
       api.fs.stat(data.storagePath, (err, stats) => {
         if (!err) data.storageSize = stats.size;
```
aqrln: No error handling or even logging.

aqrln: (Not related to this PR, though. Should be done as a separate commit.)

tshemsedinov:
Landed in 9102089

PR #714, opened 2017-03-31
Isolate api call wrapping into closure with connection

tshemsedinov:
It looks like dirty hack, but node.js do exactly same thing


lib/impress.jstp.js
```diff
@@ -63,8 +63,7 @@ impress.jstp.mixin = (application) => {
 
     const startTime = process.hrtime();
 
-    application.sandbox.connection = connection;
-    method(...args.concat([onComplete]));
+    method(connection)(...args.concat([onComplete]));
```
tshemsedinov: @aqrln Here you can inject connection

lib/impress.application.js
```diff
@@ -98,29 +98,31 @@ impress.application.mixin = (application) => {
     path.substr(application.dir.length)
   );
 
+  function apiWrap(source) {
+    return impress.USE_STRICT + '(connection => (' + source + '))';
```
tshemsedinov: @aqrln wrapping here

applications/example/config/passport.js
```diff
@@ -1,40 +0,0 @@
-if (api.passport) {
```
aqrln: Unrelated to this PR, must be a separate commit.

applications/example/lib/application.test.js
```diff
@@ -0,0 +1,31 @@
+// Example: application internal API (accessible from any place on application)
```
aqrln: Ditto.

lib/impress.application.js
```diff
@@ -187,7 +189,7 @@ impress.application.mixin = (application) => {
   }
 
   application.createSandbox = (callback) => {
-    const sandbox = { module: {}, api: {} };
```
aqrln: I guess the deprecation of `__filename`, `__dirname` and `module` should be documented in, at least, commit message.

applications/example/config/passport.js
tshemsedinov: It's unrelated but will not work after commit merged, ok, I'll do 2 commits in this PR

lib/impress.constants.js
```diff
@@ -21,7 +21,7 @@ impress.FILE_PARSE_ERROR = 13;
 impress.DIR_EXISTS = 50;
 impress.DIR_NOT_EXISTS = 51;
 
-impress.USE_STRICT = '"use strict";\n';
+impress.USE_STRICT = '\'use strict\';\n';
```
aqrln: Unrelated style change, but I'm fine with it being in this PR. I'd prefer it to be a second commit nevertheless, though.

applications/example/config/passport.js
aqrln: Ah, I see. Well, then I'm fine with it in this commit.

aqrln: But the rationale of this change should be described in the full commit message.

lib/impress.jstp.js
aqrln: Clever trick 👍

lib/impress.application.js
tshemsedinov: Yes, it deprecates `module.exports` too because we have different modularity style, so I think no need to support node.js style or strange things like `__filename`, `__dirname` (may affect security too) and because applied scripts are supposed to know where they placed.

lib/impress.constants.js
tshemsedinov: Maybe even ``` `'use strict';\n` ```

aqrln: SGTM if the linter won't complain.

tshemsedinov:
Landed in 6516b523657311f01badcfe9c5bb36fcba789bee and bedac5952920ae835870f6e06381929c0e9c5754

PR #715, opened 2017-03-31
Fixed application.log.open to prevent double callback

tshemsedinov:
Critical fix to be also included in `0.4.16`


tshemsedinov:
Landed in 0148626

PR #716, opened 2017-04-01
Isolate api call, patch for Impress v0.3.x

tshemsedinov:
- Wrap source with `codeWrap` in `application.prepareScript`
- Stop using `application.sandbox` for critical references
- Use closure to pass critical references into modules
- Deprecate: `module`, `module.exports`, `__filename` and `__dirname`

PR-URL: #714
Related commit to master: https://github.com/metarhia/Impress/commit/01486260cd93e0b958a31d9c075932499549cac3


aqrln:
Just for the record: it should work with Node.js 4.x with the "metarhia-jstp" field in the dependencies changed from ">=0.5.0" to "~0.5" or "0.5.x".

tshemsedinov:
@aqrln I know but there is no need to support `Node.js 4.x` I suppose

tshemsedinov:
Applied patch in 9de6374

aqrln:
@tshemsedinov yeah, no need *for us*. But dropping support for two release lines of Node in a *semver-patch* release is very irresponsible. I don't know, you often speak about popularizing the Metarhia projects, but being a third-party person who has ran over some project I have recently read or heard of, I would never use it if I saw its maintainers to allow themselves doing such things.

tshemsedinov:
Upgrade to `6.x` or die :skull_and_crossbones:

aqrln:
Ignoring semver compatibility guarantees isn't the best way to enforce users to upgrade.

tshemsedinov:
It was already broken on `Node 4.x` because JSTP did not incremented version on breaking changes?

aqrln:
No, it has already been broken because of invalid version range in `package.json` of Impress, as [I said above](https://github.com/metarhia/Impress/pull/716#issuecomment-290978991). It should have used either the `~` specifier (which, unsurprisingly, means "compatible with") or the explicit "0.5.x" notation. JSTP does its best to maintain compatibility in a clear and [documented](https://github.com/metarhia/JSTP/blob/master/doc/VERSIONING.md) way, but there's nothing we can do if npm isn't used properly. That, and a fact it was a critical bugfix that we had to apply ASAP, was why I approved *this* patch, but noted we should fix `package.json` (in a separate commit, probably) and didn't request changes blocking the PR until the `.travis.yml` change would have been reverted. But then you stated that it is okay and we don't need to support Node.js older than v6.0 and this thread started.

belochub:
You deprecated Node.js versions before 6.0 without actually changing `package.json`, which states that this project can be used with Node.js `>=4.0.0` versions.

belochub:
Also, next time you request a review, care to wait for it to happen?
You should not just land it as soon as someone else reviews it.

tshemsedinov:
Sorry, you are right, I removed broken `impress@0.3.37` from npm and will replace it with impress@0.3.38 after CI

PR #718, opened 2017-04-05
Metasync2 migrate

tshemsedinov:


lib/impress.files.js
```diff
@@ -69,19 +66,18 @@ Client.prototype.upload = function(
...
     doneCount++;
     if (each) each(err, data);
-    if (fileCount === doneCount && done) done(null, doneCount);
+    if (fileCount === doneCount) done(null, doneCount);
   };
 
-  if (!client.files) {
-    if (done) return done(null, 0);
-  }
+  if (!client.files) return done(null, 0);
```
agil: Can be moved to the function beginning as "guard precondition" + avoid redundant variables declarations

tshemsedinov: Moved, thanks

tshemsedinov:
@agil I use both: `function` adding methods to prototype (to access `this`) and lambdas for passing functional refs in/out of higher order functions.

lib/impress.log.js
```diff
@@ -161,14 +159,13 @@ impress.log.mixin = (application) => {
           now.getUTCFullYear(), now.getUTCMonth(), now.getUTCDate(), 0, 0, 0, 0
         );
         const time = date.getTime();
+        const cb = api.common.emptiness;
```
aqrln: This change seems to be a bit unrelated given the current commit message, but I'd rather change the latter than decouple it into a separate commit :)

How about something like this?

```
Refactor and simplify callbacks

* Use `api.metasync.cb()` callback wrapper throughout the codebase.
* Refactor `api.common.emptiness` usage for readability in
  `impress.log.js`.
```

tshemsedinov: Done

tshemsedinov:
Landed in d5316f517f16362a3147dc77ff05d4cf185ec9f7 and d5316f517f16362a3147dc77ff05d4cf185ec9f7

PR #720, opened 2017-04-07
Version 0.4.17

tshemsedinov:


tshemsedinov:
@aqrln Only critical change is `metasync2.0`. Impress use it internally and provides for applications, so if somebody uses `metasync.filter` or `metasync.find` they changed signatures. Is it ok just to write version changes?

tshemsedinov:
Landed in be50cf63f2db7c1ca374ab94773ddf99348d30c9

PR #760, opened 2017-08-20
Avoid turning on impress autostart twice

belochub:
Autostart is already being turned on during the installation of `impress-cli` package.


tshemsedinov:
Landed in 24987c4

PR #778, opened 2017-09-07
Update eslint config

aqrln:
###### First commit

This commit removes rules from `.eslintrc.yml` in favor of using the
`eslint-config-metarhia` package.  Three rules are turned off for now,
because they trigger lots of linter errors:

 - `arrow-parens` — forces consistent parens usage in arrow functions
 - `comma-dangle` — forces VCS- and diff-friendly comma usage
 - `handle-callback-err` — ensures the errors are handled instead of
   being silently ignored

How these exceptions are handled further is outside the scope of this
commit/pull request.  Former two of them are purely stylistic, but the
latter one points at legitimate potential bugs.

`eslint` package has been updated to match what `peerDependencies` of
`eslint-config-metarhia` require, and `eslint-plugin-import`, which is
also a part of the config's `peerDependencies`, has been installed.
Consequently, `eslint-plugin-impress` has been updated too, so that npm
does not complain at unmet peer dependencies (technically, it isn't
strongly connected with this patch, since eslint@4 has already been used
on `master`, so it must have been complaining already, but given that
the whole eslint infra is updated here, it kind of fits and is in spirit
of the changes).

###### Second commit

Take it or drop it, it doesn't really matter. Similar changes in other
repos involved migration to JSON for ESLint configuration files, so I
have done it here too.


tshemsedinov:
Landed in 6a13521

PR #779, opened 2017-09-16
Version 0.4.27

tshemsedinov:
- Update dependencies
- Update deploy scripts
- Update TravisCI config


tshemsedinov:
Build fails: Node 6.x on OSX
@aqrln please see https://travis-ci.org/metarhia/impress/jobs/276353822

aqrln:
I think it was fixed in https://github.com/metarhia/jstp/commit/5ed6bf0755ecd6d8308dcd84688549b6addcc31c, but that commit has never been released in neither `metarhia-jstp@0.5` nor `metarhia-jstp@0.6`. I'll check it out which commits don't have too many merge conflicts with release branches and are reasonably backportable or contain critical bugfixes, and cut new releases for both release lines.

belochub:
@aqrln, how about we just release the 1.0 version of JSTP already instead of backporting?
I think that it may even be much easier this way, there is just a couple of PRs left to finish the work on the first major version of JSTP.

tshemsedinov:
This PR doesn't change JSTP version `^0.6.8` so I can't understand failure reasons. Maybe we must remove osx build? @aqrln

aqrln:
@tshemsedinov JSTP 0.6 has never been compiling successfully on macOS with Node.js 6.0, it's just that we didn't know about it because we had Linux-only CI back when the last 0.6 was released, and no one has backported neither Travis config changes nor the patch in question to 0.6 since then.

@belochub I dunno, backporting a couple of conflict-free patches sounds an easy and profitable thing to do anyway, right? I mean, the bug is currently present in both release lines used by us, and the fix has been in `master` since a long time ago, so why not backport it? Especially given the issue; telling a dependent package's maintainers "oh, right, just update to the next major and everything's gonna work" is not that friendly semver-wise, even if they sit in one room with you.

belochub:
@aqrln, I'd like to disagree, we don't use both release lines because TLS transport is broken in version 0.6 at the moment (https://github.com/metarhia/jstp/pull/197 was never backported) and it was broken for such a long time (for almost a year now) that I don't think we have even a single project running JSTP 0.6, so I don't see much profit in updating it.

Also, I forgot to mention that after publishing 1.0 we will for sure have to update the dependents maintained by our organization, in this https://github.com/metarhia/impress/pull/779#issuecomment-330348038 I didn't mean that we should just release 1.0 and leave integration of it for someone else "sitting in one room with us".
According to npm, there are 2 dependent packages for `metarhia-jstp`, which are `impress` and `globalstorage`, and AFAIR the latter one imports the package just for serialization/deserialization functionality. This parts had no changes in the API, meaning we will only need to update `impress` after releasing the next major version for JSTP.

I am not really happy with the fact that we've been working on releasing the 1.0 for JSTP for more than half a year now and haven't been backporting the fixes, hoping they will make it in the next release, but now you want to prioritize that instead of releasing the new version, and I think it will only lead to us postponing the 1.0 release even more (even though it's been postponed so many times already). In my opinion, we should release 1.0 ASAP and use it in our projects when possible instead of using broken 0.6 or less stable 0.5 release branches. Also, I guess, afterward we can discontinue support for these versions due to them not being used in any relevant or actual projects, but if there are any proofs of me being wrong about it, I'd like you to correct me.

tshemsedinov:
Landed in d28f7b5

PR #796, opened 2018-01-05
Add possibility to write tests for applications

nechaido:


lib/impress.application.js
```diff
@@ -575,6 +586,96 @@ impress.application.mixin = (application) => {
     }
   };
 
+  if (impress.mode === 'test') {
+
+    application.loadTests = (callback) => {
+      const relPath = '/test/';
+      const absPath = application.dir + relPath;
```
belochub: You should use `path.join` instead of just concatenating here.

lib/impress.application.js
```diff
@@ -575,6 +586,96 @@ impress.application.mixin = (application) => {
...
+          callback();
+          return;
+        }
+        if (files.length === 0) {
+          callback(null);
+          return;
+        }
+        application.tests = [];
+        api.metasync.each(files, (fileName, cb) => {
+          const filePath = absPath + fileName;
```
belochub: You should use `path.join` instead of just concatenating here.

lib/impress.application.js
```diff
@@ -575,6 +586,96 @@ impress.application.mixin = (application) => {
...
+        application.tests[testName] = test;
+        callback();
+      });
+    };
+
+    application.createTest = (fileName, callback) => {
+      api.fs.readFile(fileName, (err, source) => {
+        if (err) {
+          application.log.error(impress.CANT_READ_FILE + fileName);
+          callback(err);
```
belochub: You should just return after this line to avoid excessive indentation on lines 646-653.

lib/impress.application.js
```diff
@@ -575,6 +586,96 @@ impress.application.mixin = (application) => {
...
+      });
+    };
+
+    application.runTests = (callback) => {
+      const runTest = (testName, callback) => {
+        api.tap.test(`Testing ${testName}`, (test) => {
+          application.tests[testName](test, () => {
+            test.done();
+            callback();
+            return;
```
belochub: There is no need to use `return` here.

lib/impress.application.js
```diff
@@ -575,6 +586,96 @@ impress.application.mixin = (application) => {
...
+      });
+    };
+
+    application.createTest = (fileName, callback) => {
+      api.fs.readFile(fileName, (err, source) => {
+        if (err) {
+          application.log.error(impress.CANT_READ_FILE + fileName);
+          callback(err);
+        } else {
+          const options = {
```
belochub: Since `USE_STRICT` is equal to `'\'use strict\';\n'`, and `testWrap` function puts it before the code, I'd recommend to add `lineOffset` of 1 to options for more accurate results.

lib/impress.application.js
```diff
@@ -575,6 +586,94 @@ impress.application.mixin = (application) => {
...
   };
 
+  if (impress.mode === 'test') {
+    application.loadTests = (callback) => {
+      const relPath = '/test/';
+      const absPath = api.path.join(application.dir, relPath);
+      api.fs.readdir(absPath, (err, files) => {
+        if (err) {
+          application.log.error(impress.CANT_READ_DIR + absPath);
+          callback(err);
```
belochub: Are you sure that it's needed to pass `err` into `callback` here?
Its value is ignored afterward anyway since it's only checked for existence, and this only leads to `impress` loading process being stopped, which makes this step really inconsistent with the behavior of `impress` when it encounters errors in other steps during the loading process.

lib/impress.application.js
```diff
@@ -71,19 +71,26 @@ impress.application.mixin = (application) => {
...
-          application.loadPlaces(() => {
-            application.loadApi(() => {
-              application.ready = true;
-              application.emit('started');
-              callback();
-            });
-          });
-        });
-      });
+    const fns = loadSteps.map(fn => (...args) => application[fn](...args));
```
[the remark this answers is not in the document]
nechaido: If I am correct the result will not be binded to `application`.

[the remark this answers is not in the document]
belochub: And the value of `this` isn't used in any of the functions, which means it makes no difference.

tshemsedinov:
PR paused @nechaido @belochub

tshemsedinov:
We will not use tap for Metarhia stack

PR #798, opened 2018-01-24
Update jstp

nechaido:


lib/.tern-port
```diff
@@ -0,0 +1 @@
+42355
```
belochub: What is this?

tshemsedinov:
Сhanged impress version to 0.5.0 removed lambda from frontend js

lib/.tern-port
tshemsedinov: @nechaido I don't understand this number too )

tshemsedinov:
@nechaido please see my changes

lib/.tern-port
nechaido: @tshemsedinov this file was committed by mistake, it is no longer there.

nechaido:
@tshemsedinov I don't like the idea to bump package and dependencies versions in this PR. I'd prefer to do this in a follow-up PR. I'm ok with other changes.

nechaido:
This PR lacks logging, this will be added in a corresponding commit in a few days.

tshemsedinov:
@nechaido ok, I am going to start impress 0.5.* branch after landing jstp 1.0 and leave 0.4.* with jstp 0.6.*

lib/impress.jstp.js
```diff
@@ -7,23 +7,34 @@ impress.jstp.mixImpress = true;
...
-  let apps, auth;
   if (config.name === 'master') {
-    apps = { impress };
-    auth = impress.cloud.startSession;
+    impress.version = impress.version || '1.0.0';
+    config.applications = [ impress ];
+    config.authPolicy = (
+      connection, application, strategy, [nodeId, cloudAccessKey], callback
+    ) => {
+      impress.cloud.startSession(
```
belochub: You should also change this method because now it passes generated `sessionId` to the callback as a second argument, while `username` (which in this case will probably be just `nodeId`) must be passed instead.

lib/impress.jstp.js
```diff
@@ -7,23 +7,34 @@ impress.jstp.mixImpress = true;
...
+        connection, application, nodeId, cloudAccessKey, callback
+      );
+    };
   } else {
-    apps = impress.applications;
-    auth = impress.jstp.startSession;
+    config.applications = Object.keys(impress.applications)
```
belochub: The part seems to be repeated, should it maybe be refactored into a function? 
```javascript
Object.keys(impress.applications)
      .map((key) => {
        const application = impress.applications[key];
        application.version = application.version || '1.0.0';
        console.log(application);
        return application;
      });
```

applications/example/static/js/init.js
```diff
@@ -174,34 +174,29 @@ api.dom.on('load', function() {
...
-        button.onclick = runJstpExample;
-      };
-
-      connection.handshake('example', null, null, function(err, session) {
-        if (err) {
-          print(err);
+    api.jstp.ws.connect('example',
+      new api.jstp.Application('example', {}),
+      'ws://localhost:8000',
+      function (error, connection, session) {
```
belochub: The space after `function` keyword here looks like a typo.

nechaido:
Landed in a31f4e4  and a8e6a4a.

lib/.tern-port
aqrln: @nechaido sorry for the noise if it's not relevant anymore, but FWIW I'd suggest adding `.tern-port` to global `.gitignore` if you haven't done so yet.

PR #803, opened 2018-02-25
Version 0.5.1

tshemsedinov:
Just update all dependencies deeply

### DIFF
--- a/lib/api.definition.js
+++ b/lib/api.definition.js
@@ -59,7 +59,7 @@ api.definition.walkRecursive = (
     }
   } else {
     const tplKeys = typeof(tpl) === 'object' ? Object.keys(tpl) : [];
-    const dataKeys = (
+    let dataKeys = (
       (typeof(data) === 'object' && data !== null) ? Object.keys(data) : []
     );
     other = false;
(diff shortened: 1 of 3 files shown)


belochub:
@tshemsedinov, CI build fails due to change in behavior introduced in `metarhia-common` package in version 0.0.25, specifically this commit: https://github.com/metarhia/common/commit/10d2854b324e970df66900d91d0a7705c246b3ec.
Before, `merge` function was modifying the first array provided in it, setting it to the result value, as well as returning the result value. `impress` depended on exactly that behavior, and in the aforementioned patch, this function was modified to not change the input, only returning the result value instead.
I think it's the issue in the `metarhia-common` package, this kind of change in behavior must not have been introduced in a new patch version. I'd recommend fixing the `metarhia-common` package and republishing it as a new patch version. Another option is to update `impress` to make it work with this new behavior, but I don't think it is a good idea since it was not an issue in `impress` in the first place and it may look like encouraging the introduction of breaking changes in the patch versions.

tshemsedinov:
@belochub I understand that it breaks semver general approach but we have all libs versions `0.*` and it is not so strict for leading-zero versions. So I'd like to fix it in impress.

tshemsedinov:
@belochub thank you for detecting reasons of this problem

tshemsedinov:
Landed in 15e4551

aqrln:
@tshemsedinov as far as I understand, `metarhia-common` is pretty stable and most of its code has undergone years of extensive real-world testing as a part of Impress. Should we maybe consider bumping it to 1.0?

PR #812, opened 2018-04-04
Use metalog instead of builtin logger

tshemsedinov:
Closes: https://github.com/metarhia/impress/issues/810


tshemsedinov:
Landed in e9cf327 and d6750b2

PR #822, opened 2018-04-16
Add missing functionality to console emulation

belochub:
This implementation fully mimics original API.

I am not completely sure about some of the styling, but from the functionality point of view, this seems to be working correctly.


belochub:
Landed in https://github.com/metarhia/impress/commit/ff26474cccfffac6ea625350cf94e8dd8656dd57.

PR #848, opened 2018-07-05
Avoid creation of redundant fields in application

belochub:
After the introduction of the plugin system, `application` object
included some unwanted fields named after the plugins with values
set to an empty object.


belochub:
Landed in https://github.com/metarhia/impress/commit/f2fa99d7da18774571ba351d5aa5195452f3df3d.

PR #850, opened 2018-07-07
 Remove tets stubs and unittests for metarhia-common

tshemsedinov:


tshemsedinov:
Landed in f0fdb9c and af5cc1e

PR #855, opened 2018-07-24
Use argon2 for password hashing

nechaido:


lib/security.js
```diff
@@ -96,9 +96,16 @@ const mixin = (application) => {
         callback(new Error(AUTH_ERROR));
         return;
       }
-      const passwordHash = security.hash(password);
-      if (passwordHash === user.password) callback(null, user);
-      else callback(new Error(AUTH_ERROR));
+
+      api.argon2.verify(user.password, password).then((matches, error) => {
```
belochub: Same here, you should use `Promise.catch`.

lib/security.js
```diff
@@ -128,19 +135,24 @@ const mixin = (application) => {
     password, // string
     callback // function(err, user)
   ) => {
-    const passwordHash = security.hash(password);
-    security.getUser(login, (err, user) => {
-      if (user) {
-        callback(new Error('Login already registered'), user);
+    security.hash(password, (hash, error) => {
```
belochub: In `security.hash` you are passing an `error` into callback as a first argument, and `hash` as a second argument.

lib/security.js
```diff
@@ -75,15 +75,15 @@ const mixin = (application) => {
...
-    const salt = application.config.sessions.secret;
-    const result = api.crypto
-      .createHmac('sha512', salt)
-      .update(password)
-      .digest('latin1');
-    return result;
+    api.argon2.hash(
+      password,
+      { type: api.argon2.argon2id }
+    ).then((hash, error) => {
```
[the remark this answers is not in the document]
tshemsedinov: I thought it had such a strange contract, so we need a twist function `(hash, error) => callback(error, hash)` @belochub and I believe @nechaido knows then/catch contract.

tshemsedinov:
@nechaido do they have callback interface?

nechaido:
@tshemsedinov, no, they do not.

nechaido:
Landed in 1fcac69.

PR #856, opened 2018-07-28
Update fs usages according to node v10 deprecations

lundibundi:
As per this https://github.com/nodejs/node/pull/12562 node PR callback
in async fs functions is no longer optional. This commit fixes
remaining usages in impress.

I've scanned impress, impress-cli and globalstorage for any of the usages of those fs functions. This should be the last one.


tshemsedinov:
Landed in fa3bc75

PR #866, opened 2018-08-29
Refactor loop variables

JuliaGerasymenko:


lib/api.definition.js
```diff
@@ -274,9 +274,9 @@ api.definition.printErrors = (
   if (validationResult.valid) return;
 
   impress.log.error(message);
-  let i, len, error;
-  for (i = 0, len = validationResult.errors.length; i < len; i++) {
-    error = validationResult.errors[i];
+  const len = validationResult.errors.length;
```
nechaido: You don't need to cache array length.

lib/application.js
```diff
@@ -113,11 +113,11 @@ const mixin = (application) => {
   };
 
   application.systemTemplates = {};
-  let j, jlen, tplName, tplFile, tplPath;
-  for (j = 0, jlen = TEMPLATES.length; j < jlen; j++) {
-    tplName = TEMPLATES[j];
-    tplFile = '/templates/' + tplName + '.template';
-    tplPath = application.dir + tplFile;
+  const jlen = TEMPLATES.length;
```
nechaido: You don't need to cache array length.

lib/application.js
```diff
@@ -232,23 +232,24 @@ const mixin = (application) => {
       const globals = (
         application.config.sandbox.global || DEFAULT_SANDBOX
       );
-      let i, len, moduleLink, moduleName, msg;
-      for (i = 0, len = globals.length; i < len; i++) {
-        moduleName = globals[i];
+      const len = globals.length;
```
nechaido: Ditto.

lib/application.js
```diff
@@ -232,23 +232,24 @@ const mixin = (application) => {
...
             moduleLink = api.registry.deprecate(moduleLink, msg);
           }
           application.sandbox[moduleName] = moduleLink;
         }
       }
       const apis = application.config.sandbox.api || api.registry.defaultNames;
-      let j, jlen;
-      for (j = 0, jlen = apis.length; j < jlen; j++) {
-        moduleName = apis[j];
+      const jlen = apis.length;
```
nechaido: Ditto.

lib/application.js
```diff
@@ -512,11 +514,11 @@ const mixin = (application) => {
     if (!application.isInitialized && !application.isImpress) {
       application.isInitialized = true;
       if (config.routes) { // Prepare application routes
-        let j, jlen, route, rx;
         const routes = config.routes;
-        for (j = 0, jlen = routes.length; j < jlen; j++) {
-          route = routes[j];
-          rx = !route.escaping ? route.url : ('^' + route.url
+        const jlen = routes.length;
```
nechaido: Ditto.

lib/cache.js
```diff
@@ -164,8 +162,8 @@ const mixin = (application) => {
         cb();
       });
     }, () => {
-      let i, ilen;
-      for (i = 0, ilen = files.length; i < ilen; i++) {
+      const ilen = files.length;
```
nechaido: Ditto.

lib/application.js
```diff
@@ -689,10 +691,10 @@ const mixin = (application) => {
 
     const routes = application.config.routes;
     if (routes) {
-      let i, len, route, match;
-      for (i = 0, len = routes.length; i < len; i++) {
-        route = routes[i];
-        match = req.url.match(route.urlRx);
+      const len = routes.length;
```
nechaido: Ditto.

lib/api.db.mysql.js
```diff
@@ -107,33 +107,28 @@ if (api.mysql) {
...
-          fieldIndex = 1;
-          indexes = [];
-          references = [];
+          const table = databaseSchema[tableName];
+          const primaryKey = [];
+          const indexes = [];
+          const references = [];
+          let referenceStatement, reference;
+          let fieldsDef = '';
+          let primaryFieldDef = '';
```
nechaido: This variable may be declared later, before its first actual initialization.

lib/application.js
```diff
@@ -491,17 +492,18 @@ const mixin = (application) => {
     }
 
     const servers = config.servers;
-    let server, serverName, serviceName, cpus, i, ilen, srv, port;
     config.servers = {};
-    for (serverName in servers) {
-      server = servers[serverName];
+    for (const serverName in servers) {
+      const server = servers[serverName];
+      let serviceName;
```
belochub: This declaration should probably be moved to line 508 since it is not used anywhere else, you can also use `const` there.

lib/application.js
```diff
@@ -232,23 +232,24 @@ const mixin = (application) => {
...
           application.sandbox[moduleName] = moduleLink;
         }
       }
       const apis = application.config.sandbox.api || api.registry.defaultNames;
-      let j, jlen;
-      for (j = 0, jlen = apis.length; j < jlen; j++) {
-        moduleName = apis[j];
+      const jlen = apis.length;
+      for (let j = 0; j < jlen; j++) {
+        let moduleLink, moduleName = apis[j];
```
belochub: `moduleLink` declaration should be moved to the next line, where it is also being initialized.

lib/application.js
JuliaGerasymenko: I did as you wanted, but the line 508 has the maximum line length of 80.

nechaido:
```js
    const serviceName =
      serverName === "master" ? serverName : serverName + port;
```

lib/health.js
```diff
@@ -155,16 +154,18 @@ const cmdExec = (
...
-    for (i = 0, ilen = lines.length; i < ilen; i++) {
-      line = lines[i];
+    const ilen = lines.length;
+    for (let i = 0; i < ilen; i++) {
+      let header, col, row;
+      let line = lines[i];
       if (line) {
         line = line.replace('Use%', 'Use').replace('Mounted on', 'MountedOn');
         col = line.split(process.isWin ? rxCSV : rxSPC);
         if (!header) header = col;
```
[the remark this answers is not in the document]
JuliaGerasymenko: Sorry, what did you mean?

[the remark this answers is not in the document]
belochub: There was a mistake in my comment above, what I wanted to say is the other branch (`else` branch) of this `if` statement will never be executed because you are declaring new `header` variable on each loop iteration and there is no value assigned to it before this point, meaning that on this line `header` is always equal to `undefined`. 

That being said, this code won't work in the same way as before.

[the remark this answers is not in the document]
JuliaGerasymenko: will it work if I declare a header before the loop?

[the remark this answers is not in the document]
JuliaGerasymenko: What I need to do that code will work again?

lib/health.js
```diff
@@ -155,16 +153,16 @@ const cmdExec = (
       return;
     }
     const lines = res.split(rxLine);
-    let i, ilen, j, jlen, line, header, col, row;
-    for (i = 0, ilen = lines.length; i < ilen; i++) {
-      line = lines[i];
+    for (let i = 0; i < lines.length; i++) {
+      let header;
```
nechaido: I believe `header` must be declared outside of the loop.

tshemsedinov:
I afraid we need do it manually @JuliaGerasymenko

JuliaGerasymenko:
I have some errors even if I'm on the branch master
image

belochub:
Landed in https://github.com/metarhia/impress/commit/5b7b832ba8f5738bb38a0a40cb6a5e48bc9f0513.

PR #873, opened 2018-10-05
Version 0.7.0

tshemsedinov:


package.json
```diff
@@ -49,31 +49,31 @@
...
-    "@metarhia/jstp": "^2.0.0",
+    "@metarhia/jstp": "^2.1.0",
     "accept-language": "^3.0.18",
     "argon2": "^0.19.3",
-    "concolor": "^0.1.8",
-    "csv": "^2.0.0",
-    "globalstorage": "^0.2.6",
+    "concolor": "^0.1.9",
+    "csv": "^4.0.0",
+    "globalstorage": "^0.2.8",
```
belochub: What do you think about upgrading `globalstorage` to `^0.3.0`, since this is a minor version upgrade for `impress`?

tshemsedinov: This will remove mongodb support. So we need more changes in impress to be ready for this.

tshemsedinov: `lib/security.js` expects that we have db with globalstorage API here `application.database.security`

belochub: Okay, I see, let's release the v0.7 like this then, and afterward, we should probably start updating the security module so that we can update `impress` to use the newer versions of `globalstorage`.

Can we maybe assign @o-rumiantsev, @JuliaGerasymenko or @bugagashenkj on this task?

tshemsedinov:
Landed in 2eea908

PR #875, opened 2018-10-18
WIP: Use metaschema for config validation

o-rumiantsev:
I have a question about how to load schemas. Using synchronous `api.definition.require` it is loading schemas when `lib/core.js` is required
https://github.com/metarhia/impress/blob/2eea90822b682e7e1195376c3f8ecdcfed85ab1e/lib/core.js#L52-L60

But since the `api.metaschema.load` works asynchronously I think loading schemas on require is a bad idea
https://github.com/metarhia/impress/blob/1feab16b0ff6be07cc8dec46862801649c92f437/lib/core.js#L52-L76

I propose to define a separate function `impress.loadSchemas` and call it before `impress.loadConfig`
https://github.com/metarhia/impress/blob/2eea90822b682e7e1195376c3f8ecdcfed85ab1e/lib/core.js#L233-L267

Also there is some problems with validation. For example:
In `config.application.definition` there is a schema for `application` config, where field `slowTime` defined with a type `duration`
https://github.com/metarhia/impress/blob/2eea90822b682e7e1195376c3f8ecdcfed85ab1e/schemas/config.application.definition.js#L15-L20

And in `applications/example/config/application` `slowTime` is set as `'2s'`(string)
https://github.com/metarhia/impress/blob/2eea90822b682e7e1195376c3f8ecdcfed85ab1e/applications/example/config/application.js#L4

But in `metaschema` domain `Duration` has type `number`. What should I do then @tshemsedinov @nechaido @belochub?

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -26,8 +26,22 @@ class Config {
 
   constructor(application) {
     this.application = application;
+    this.schemas = null;
     this.sandbox = null;
     this.sections = {};
+  }
+
+  loadSchemas(callback) {
+    const schemasDir = api.path.join(impress.moduleDir, 'schemas/config');
+    api.metaschema.fs.loadAndCreate(schemasDir, { api }, (err, schemas) => {
+      if (err) {
+        impress.log.error(err);
+        callback();
+        return;
+      }
+      this.schemas = schemas;
+      callback();
+    });
   }
 
   loadSandboxConfig(callback) {
@@ -91,6 +105,10 @@ class Config {
         return;
       }
       this.sections[sectionName] = exports;
+      if (this.schemas.categories.has(sectionName)) {
+        const error = this.schemas.validateCategory(sectionName, exports);
+        if (error && impress.isMaster) console.error(error.toString());
+      }
       callback();
     });
   }


tshemsedinov:
I will refactor according to #879 and after that we will discuss your code @o-rumiantsev

lib/api.registry.js
```diff
@@ -157,6 +158,7 @@ api.registry.load = () => {
   api.common = require('metarhia-common');
   api.sandboxedFs = require('sandboxed-fs');
   api.json = JSON;
+  api.metaschemaLoader = require('metaschema');
```
tshemsedinov: Explain this changes please

schemas/config.impress.definition/domains.schema
```diff
@@ -0,0 +1,11 @@
+{
+  Instance: Enum('standalone', 'controller', 'server'),
+
+  Nomens: Many({ domain: 'Nomen' }),
```
o-rumiantsev: @tshemsedinov I wonder whether I can do this, if I need to define, that the value of this domain can be an array of strings?

lib/core.js
```diff
@@ -53,6 +53,32 @@ impress.applicationConfigDefinition = api.definition.require(
...
+  api.metaschema.fs.loadAndCreate(
+    [serverConfigSchemas, applicationConfigSchemas],
+    api,
+    (err, metaschema) => {
+      if (err) {
+        console.error(err);
+        return;
+      }
+
+      api.metaschema = metaschema;
```
o-rumiantsev: I am not sure this is the right decision to reassign `api.metaschema`. Maybe it is better to name it `api.ms`?

schemas/config.application.definition/database.schema
```diff
@@ -0,0 +1,7 @@
+{
+  alias: { domain: 'Nomen' },
+  url: { domain: 'URL', required: true },
+  collections: { domain: 'Nomens' },
+  slowTime: { domain: 'Duration', required: true },
+  security: { domain: 'Logical', required: true }
```
lundibundi: there is also `poolMax` in oracleConnection now.

tshemsedinov:
This will be implemented in a different way in Impress 2.x
I'll close this PR but don't delete branch

Issue #878, opened 2018-10-24
Remove mongodb and add pg in deploy script


tshemsedinov:
Remove `*-mongo.sh` https://github.com/metarhia/impress/tree/master/deploy
Add `*-pgsql.sh` for all operation systems

@nechaido I propose this task for @o-rumiantsev

belochub:
@tshemsedinov, this issue's title contradicts the description, probably a typo:
>Remove mongodb and **add** pg in deploy script
>**Remove** `*-pgsql.sh` for all operation systems

tshemsedinov:
@belochub thanks, fixed

PR #880, opened 2018-11-08
Add deploy scripts for PostgreSQL

o-rumiantsev:
Closes: https://github.com/metarhia/impress/issues/878


deploy/centos7x64-pgsql.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+sudo rpm -Uvh https://yum.postgresql.org/11/redhat/rhel-7-x86_64/pgdg-centos11-11-2.noarch.rpm
+sudo yum install -y postgresql11 postgresql11-server postgresql11-contrib
+sudo /usr/pgsql-11/bin/postgresql-11-setup initdb
+sudo systemctl enable postgresql-11
+sudo systemctl start postgresql-11
```
lundibundi:
```suggestion
sudo systemctl enable --now postgresql-11
```
For both enable and start.

deploy/fedora-pgsql.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
-sudo dnf -y install postgresql-server postgresql-contrib
+sudo dnf -y install postgresql postgresql-server postgresql-contrib
 sudo systemctl enable postgresql
+sudo postgresql-setup --initdb --unit postgresql
```
lundibundi: Maybe move this one line above and use `enable --now`?

deploy/ubuntu-pgsql.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+sudo apt update;
+sudo apt install -y postgresql postgresql-contrib
+cd /ias
```
lundibundi: Why don't you enable/start PG anymore?

lundibundi: Could you add a comment about that to avoid possible confusion during further upgrades of those files?

deploy/debian-pgsql.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+RELEASE=$(lsb_release -sc)
+sudo sh -c "echo \"deb http://apt.postgresql.org/pub/repos/apt/ $RELEASE-pgdg main\" >> /etc/apt/sources.list.d/pgdg.list"
+sudo apt update
+sudo apt install -y --force-yes postgresql postgresql-contrib
```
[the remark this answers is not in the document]
o-rumiantsev: The problem happens on Debian 8
```
WARNING: The following packages cannot be authenticated!
  libpq5 pgdg-keyring postgresql-client-common postgresql-client-11 postgresql-common postgresql-11 postgresql postgresql-contrib
E: There are problems and -y was used without --force-yes
```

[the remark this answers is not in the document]
o-rumiantsev: The problem happens on Debian 8
```
WARNING: The following packages cannot be authenticated!
  libpq5 pgdg-keyring postgresql-client-common postgresql-client-11 postgresql-common postgresql-11 postgresql postgresql-contrib
E: There are problems and -y was used without --force-yes
```

People are trying to solve this with `sudo apt-key update`, but it doesn't help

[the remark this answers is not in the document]
belochub: Have you tried following the instruction [here](https://wiki.postgresql.org/wiki/Apt)?
Maybe importing the repository key from https://www.postgresql.org/media/keys/ACCC4CF8.asc can help:
```
sudo apt-get install curl ca-certificates
curl https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add -
```

tshemsedinov:
Landed in 9f7ec88

Issue #889, opened 2018-11-10
Support IPv6 addresses


tshemsedinov:
Now we convert IPv4 to IPv6 removing prefix `::ffff:` but we need to convert addresses to `number` and later to `BigInt`

tshemsedinov:
It works after we reworked impress to 2.x

PR #935, opened 2018-11-27
Review to improve v1.x

tshemsedinov:


lib/cache.js
```diff
@@ -2,6 +2,8 @@
 
 // Impress Application Full Memory Cache
 
+const DEFAULT_WATCH_INTERVAL = 2000;
```
lundibundi: I think it will create a lot of unnecessary fs calls and cpu load at 2 seconds. I think we should maybe use 2000 for `test` mode but have it at maybe 1 minute in `prod`, wdyt?

tshemsedinov: This doesn't means that we will create multiple timers or check fs every 2 seconds, this just mean that updated files will not be recached immediately after receiving watch event and we will wait at least 2 seconds (by default) for more watch events to process them together. Server configuration may override this interval. For dev mode it's better to minimize this interval, for production it may be increased but not 60 sec, maybe to 10-15 sec. @lundibundi @nechaido

lundibundi: Could we then rename it to `DEFAULT_WATCH_DEBOUNCE_TIME` as that will make it a bit clearer IMO.
I'm +1 to having it 2000 for dev mode and 10000 in production.

Could we use something like
```js
const DEFAULT_WATCH_DEBOUNCE_TIME = process.env.IMPRESS_MODE === 'prod' ? 10000 : 2000;
```
?

lib/workers.js
```diff
@@ -20,26 +20,26 @@ impress.fork = (env, inspect) => {
```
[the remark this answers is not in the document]
tshemsedinov: Fixed in https://github.com/metarhia/impress/pull/967

lib/application.js
```diff
@@ -234,16 +235,23 @@ class Application extends api.events.EventEmitter {
 
   loadPlaces(callback) {
     api.metasync.each(PLACES, (placeName, cb) => {
-      this.loadPlaceScripts(placeName, cb);
+      if (placeName === 'static') {
+        cb();
+        return;
+      } else if (placeName === 'schemas') {
+        const dir = api.path.join(this.dir, 'schemas');
+        api.metaschema.fs.loadAndCreate(dir, null, (error, schemas) => {
```
lundibundi: Shouldn't we at least pass `api` as a second argument to the `loadAndCreate`?

nechaido: @tshemsedinov we cannot pass `null`, we need at least `api.common.Uint64` and `gs` available in schemas.

lib/client.js
```diff
@@ -138,12 +81,21 @@ class Client {
       res.setHeader('Keep-Alive', 'timeout=' + timeout);
     }
     if (this.ip) {
-      this.ip = ipv6toipv4(this.ip);
-      this.ipInt = api.common.ipToInt(this.ip);
+      const ip = api.net.isIPv6(this.ip) ? this.ip.slice(7) : this.ip;
```
belochub: I am not sure this will work with any IPv6 address, for example, if we get `2001:db8::ff00:42:8329`, the result of executing this line will set the `ip` value to `8::ff00:42:8329` which is an entirely different IPv6 address.
If this was supposed to be a workaround for [IPv4-Mapped IPv6 Addresses](https://tools.ietf.org/html/rfc4291#section-2.5.5.2), then we should check if it is the case for each IP address and only then extract the IPv4 out of it if needed.

tshemsedinov: We have separate issue for this: https://github.com/metarhia/impress/issues/889

lib/client.js
```diff
@@ -876,153 +872,42 @@ class Client {
...
+    const fileCache = application.cache.templates.get(relPath);
+    if (fileCache) {
+      callback(null, fileCache);
       return;
     }
-    api.fs.readFile(filePath, 'utf8', (err, tpl) => {
+    api.fs.readFile(filePath, 'utf8', (err, data) => {
       if (err) {
-        callback(null, impress.TPL_NOT_FOUND + filePath);
+        impress.log.error(impress.CANT_READ_FILE + filePath);
```
lundibundi: Shouldn't this call a `callback` too?

tshemsedinov:
Closed due to inactivity

PR #981, opened 2018-12-10
Refactor db subsystem

tshemsedinov:


tshemsedinov:
Landed in 4b6cfdd, 880388c, 7ab1b32 and 624ac53

lib/application.js
```diff
@@ -241,9 +241,11 @@ class Application extends api.events.EventEmitter {
         return;
       } else if (placeName === 'schemas') {
         const dir = api.path.join(this.dir, 'schemas');
-        api.metaschema.fs.loadAndCreate(dir, null, (error, schemas) => {
+        const provider = api.gs.provider;
+        api.metaschema.fs.loadAndCreate(dir, provider, (error, schemas) => {
```
belochub: Passing provider as a second argument here, it will be resolved to the `api` field in the `global` object inside the schema execution context.

lib/db.gs.js
```diff
@@ -14,24 +13,24 @@ if (api.gs) {
...
-      if (db) {
-        config.gs = api.gs;
-        config.db = db.connection;
-        config.provider = db.schema;
-        database.connection = api.gs;
-      }
-      api.gs.open(config, err => {
+    const config = database.config;
+    const provider = api.gs('pg', config);
+    api.gs.provider = provider;
```
belochub: Does this mean that provider instance's methods will be accessible via `api.gs.provider.methodName()`? Or is there a shorthand available, maybe something similar to `gs.methodName()`?

tshemsedinov: In applications you can use database alias (from configs)

lib/application.js
tshemsedinov: I'll pass `application.sandbox` as a second argument

PR #1001, opened 2018-12-15
Optimize config access through all submodules

tshemsedinov:
Closes: Optimize config access through all submodules


tshemsedinov:
Landed in eac4cb8

PR #1008, opened 2018-12-16
Version 1.0.0-alpha.8

tshemsedinov:


lundibundi:
Perhaps we should also run `npm update`?

Also, could you check the changes to package-lock.json as for some reason `fsevents` got removed while nothing in the `package.json` updated.

tshemsedinov:
@lundibundi I have executed `npm update` and `npm dedupe` and it removed `fsevents`

tshemsedinov:
Landed in cbb2426

tshemsedinov:
@lundibundi `fsevents` was dependency of `chokidar` but now `chokidar` doesn't depend on it. I don't know why npm didn't removed it before

PR #1029, opened 2018-12-28
Implement AsyncFunction handlers

tshemsedinov:
Closes: https://github.com/metarhia/impress/issues/951


lib/client.js
```diff
@@ -598,18 +597,23 @@ class Client {
...
       }
     };
 
-    try {
-      fn(this, done);
-    } catch (err) {
-      this.error(500, err);
-      application.logException(err);
-      callback();
+    if (fn.constructor.name === 'AsyncFunction') {
```
belochub: It would be safer to write
```suggestion
    if (Object.prototype.toString.call(fn) === '[object AsyncFunction]') {
```

lib/client.js
```diff
@@ -598,18 +597,23 @@ class Client {
...
+      try {
```
belochub: I think we can just use `Promise` API here:
```suggestion
      fn(this).then(result => {
        done(null, result);
      }, err => {
        done(err);
      });
```
This will also allow to make the method `executeFunction()` non-async, which in turn will allow to use it with Node.js versions where `async` is not yet available.

tshemsedinov:
Landed in 0b1c8b2

PR #1054, opened 2019-01-08
Fix cache purge

o-rumiantsev:


lib/cache.js
```diff
@@ -89,12 +89,12 @@ class Cache {
 
   purge() {
     if (!this.purgeNeeded()) return;
-    this.static.forEach((item, name) => {
```
belochub: What exactly was wrong with using a `forEach()` method here?

o-rumiantsev: The aim was to return from `purge`, when there is no purge needed, but this `return` makes it to leave from lambda passed to `forEach`, but not `purge`.
https://github.com/metarhia/impress/blob/04966f43638b6a9f1c11d7728f6af9e940060f78/lib/cache.js#L95

tshemsedinov:
Landed in 9eb752b

PR #1057, opened 2019-01-10
Fix start-stop events and flags

tshemsedinov:
- Added event `Application.on('stopped')`
- Added flag `Application.stopped`
- Removed event `Application.on('stop')`
- Removed event `Application.on('start')`
- Renamed flag `Application.ready` to `Application.started`
- Removed flag `Application.isInitialized`

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -543,26 +543,18 @@ impress.stop = callback => {
     return;
   }
 
-  let count = 0;
-  const afterClose = () => {
-    count--;
-    for (const appName in impress.applications) {
-      const application = impress.applications[appName];
+  const list = Object.values(servers).filter(server => server.instance);
+  api.metasync.each(list, (server, next) => {
+    server.instance.close(next);
+  }, () => {
+    for (const application of Object.values(impress.applications)) {
       application.scheduler.stopTasks();
       application.cache.clear();
       application.stopped = true;
       application.emit('stopped');
     }
-    if (count === 0) callback();
-  };
-
-  for (const serverName in servers) {
-    const server = servers[serverName];
-    if (server.instance) {
-      count++;
-      server.instance.close(afterClose);
-    }
-  }
+    callback();
+  });
 };
 
 impress.shutdown = (code = 0) => {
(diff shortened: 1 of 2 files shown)


lib/core.js
```diff
@@ -534,24 +534,34 @@ impress.gsInterval = () => {
...
+impress.stop = callback => {
   impress.cache.clear();
   const servers = impress.servers;
   if (!servers) {
     impress.log.warn('No servers active');
+    callback();
     return;
   }
+
+  let count = 0;
```
belochub: Can we maybe use `metasync.each()` here instead of manual counting?

lib/core.js
```diff
@@ -534,24 +534,34 @@ impress.gsInterval = () => {
...
+  const afterClose = () => {
+    count--;
+    for (const appName in impress.applications) {
+      const application = impress.applications[appName];
+      application.emit('stop');
```
belochub: Do we need both `stop` and `stopped` events?

tshemsedinov: 'stopped' is enough but we have 'start' and 'started' so we can remove both 'stop' and 'start' for symmetry :smile:

lib/core.js
tshemsedinov: `metasync.each()` iterates `Array` but we need to iterate `Object`, we can take keys but I think counter is ok here

belochub: As another option, we can use `Object.values()` here with `metasync.each()`.

lib/core.js
tshemsedinov: Removed

lib/core.js
```diff
@@ -534,24 +534,33 @@ impress.gsInterval = () => {
...
   if (!servers) {
     impress.log.warn('No servers active');
+    callback();
     return;
   }
+
+  let count = 0;
+  const afterClose = () => {
+    count--;
+    for (const appName in impress.applications) {
```
lundibundi: We can also use `Object.values` here I think.

lib/core.js
tshemsedinov: Ok

tshemsedinov:
Landed in 1f24800

PR #1078, opened 2019-01-23
Update package json and use new metaschema

lundibundi:


tshemsedinov:
Also I removed schema file and `main` field fix in `package.json`

tshemsedinov:
Landed in bfc9ef7

PR #1095, opened 2019-02-08
Fix save session

tshemsedinov:


tshemsedinov:
Landed in 40ceb0c and 5114e49

PR #1097, opened 2019-02-13
Update metaschema

nechaido:


lib/application.js
```diff
@@ -202,23 +202,39 @@ class Application extends api.events.EventEmitter {
...
+    const dir = api.path.join(this.dir, 'schemas');
+    const loaderPath = api.path.join(dir, 'config.js');
+    impress.createScript(this, loaderPath, (err, getMsConfig) => {
+      if (err) {
+        callback(err);
+        return;
+      }
+
+      const { options, config } = getMsConfig();
+      options.context = this;
```
lundibundi: IIUC only `sandbox` was passed previously, is this correct that we now pass whole Application?

lib/application.js
```diff
@@ -202,23 +202,39 @@ class Application extends api.events.EventEmitter {
...
+
+      api.metaschema.fs.load(dir, options, config).then(([errors, schema]) => {
+        if (errors.length) {
```
lundibundi: I'd prefer to be explicit `errors.length > 0`.

lib/application.js
```diff
@@ -202,23 +202,39 @@ class Application extends api.events.EventEmitter {
...
+          console.error(errors);
+          return;
```
lundibundi: You don't seem to call `callback` here, wouldn't this hang?

package-lock.json
```diff
@@ -466,7 +466,7 @@
         },
         "util": {
           "version": "0.10.3",
-          "resolved": "https://registry.npmjs.org/util/-/util-0.10.3.tgz",
+          "resolved": "http://registry.npmjs.org/util/-/util-0.10.3.tgz",
```
lundibundi: Revert those back to https please.

applications/example/schemas/config.js
```diff
@@ -0,0 +1,3 @@
+() => {
+  return api.metaschema.default;
+}
```
tshemsedinov: `() => api.metaschema.default;`

lib/application.js
```diff
@@ -202,23 +202,39 @@ class Application extends api.events.EventEmitter {
...
+        }
+        this.schema = schema;
+        callback(null);
+      }, error => {
+        callback(error);
+      }).catch(callback);
```
tshemsedinov:
```suggestion
      });
```
No need to `catch` here, all errors will be passed to second function in `then`

lib/application.js
```diff
@@ -202,23 +202,39 @@ class Application extends api.events.EventEmitter {
...
+        callback(err);
+        return;
+      }
+
+      const { options, config } = getMsConfig();
+      options.context = this;
+
+      api.metaschema.fs.load(dir, options, config).then(([errors, schema]) => {
+        if (errors.length) {
+          console.error(errors);
```
tshemsedinov: Use `impress.log.error()`. I think we can use `console.*()` only before `metalog` initialization.

lib/application.js
```diff
@@ -202,23 +202,39 @@ class Application extends api.events.EventEmitter {
     });
   }
 
+  loadSchemas(callback) {
+    const dir = api.path.join(this.dir, 'schemas');
+    const loaderPath = api.path.join(dir, 'config.js');
+    impress.createScript(this, loaderPath, (err, getMsConfig) => {
+      if (err) {
+        callback(err);
```
tshemsedinov: Need to write logs here

lib/application.js
nechaido: This catch will be used if `then` function throws an error.

lib/application.js
```diff
@@ -202,23 +202,42 @@ class Application extends api.events.EventEmitter {
...
+        .then(([errors, schema]) => {
+          if (errors.length > 0) {
+            this.log.error(errors);
+            callback(errors);
+            return;
+          }
+          this.schema = schema;
+          callback(null);
+        }, callback)
+        .catch(callback);
```
belochub: It is now possible that the callback will be called more than once, and if we need to keep using promises here, I think the better way of doing this would be:
```js
      api.metaschema.fs
        .load(dir, options, config)
        .then(([errors, schema]) => {
          if (errors.length > 0) {
            this.log.error(errors);
            return Promise.reject(errors);
          }
          this.schema = schema;
        })
        .then(callback, callback);
```
But in my opinion, an even better solution would be to just use `util.callbackify()` to avoid mixing two styles:
```js
      const loadSchema = api.util.callbackify(api.metaschema.fs.load);
      loadSchema(dir, options, config, (err, result) => {
        if (err) {
          callback(err);
          return;
        }
        const [errors, schema] = result;
        if (errors.length > 0) {
          this.log.error(errors);
          callback(errors);
          return;
        }
        this.schema = schema;
        callback(null);
      }
```

tshemsedinov:
@nechaido please rebase on master and I'll land this

nechaido:
@tshemsedinov done.

tshemsedinov:
Landed in ec2dc06

PR #1100, opened 2019-02-14
Load application places sequentially

nechaido:
It appears that all application places are loaded simultaneously, which looks like a bug: intuitively `utils` should be loaded before `lib`, this PR aims to fix that.

This behavior was encountered by @lundibundi.


nechaido:
@tshemsedinov please land this ASAP.

tshemsedinov:
Landed in 9a30a72

PR #1104, opened 2019-02-16
Make schemas/config.js optional

lundibundi:
This will avoid impress crash when schemas/config.js doesn't exist.

e.g. when you create Impress application to only serve http or jstp requests you may not need the schemas at all.

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -139,7 +139,10 @@ class Application extends api.events.EventEmitter {
             const path = api.path.join(relPath, fileName);
             this.preloadDirectory(path, depth - 1, next);
           } else if (fileExt === 'js') {
-            impress.createScript(this, filePath, next);
+            impress.createScript(this, filePath, (err, func) => {
+              if (err) this.log.error(impress.CANT_READ_FILE + filePath);
+              next(err, func);
+            });
           } else {
             next();
           }
@@ -196,6 +199,7 @@ class Application extends api.events.EventEmitter {
 
     impress.createScript(this, path, (err, test) => {
       if (err) {
+        this.log.error(impress.CANT_READ_FILE + path);
         callback(err);
         return;
       }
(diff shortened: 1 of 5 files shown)


lib/scripts.js
```diff
@@ -78,15 +73,37 @@ impress.createScript = (application, fileName, callback) => {
...
   });
 };
 
+// Create script if file exists
+//   application <Application>
+//   fileName <string> absolute path
+//   callback <Function>
+//     err <Error>
+//     exports <Function> | <null>
+impress.createScriptOpt = (application, fileName, callback) =>
```
tshemsedinov: We do not need additional function for optional scripts. Better to handle err-back at the higher level.

lib/application.js
```diff
@@ -340,11 +340,14 @@ class Application extends api.events.EventEmitter {
   loadSchemas(callback) {
     const dir = api.path.join(this.dir, 'schemas');
     const loaderPath = api.path.join(dir, 'config.js');
-    impress.createScript(this, loaderPath, (err, getMsConfig) => {
+    impress.createScriptOpt(this, loaderPath, (err, getMsConfig) => {
       if (err) {
```
tshemsedinov: Here we can check `err.code`

lib/scripts.js
lundibundi: But this way the error will be written to logs as `error`. IMO it will be very confusing.
Is that okay?

tshemsedinov: We need to move log writing higher too

lundibundi:
Also, there are a few places that doesn't forward error in the callback I noticed while fixing this, @tshemsedinov could you please check that it's okay to ignore error there?

tshemsedinov:
Rebased on master

tshemsedinov:
Landed in e386a9a and fa817bc

PR #1107, opened 2019-02-19
Invalid usage of metalog

JuliaGerasymenko:


lib/db.mysql.js
```diff
@@ -27,7 +27,7 @@ if (api.mysql) {
 
     connection.connect(err => {
       if (err) {
-        database.application.log.error(err);
+        database.application.log.error(err.toString());
```
SemenchenkoVitaliy:
```suggestion
        database.application.log.error(err.stack);
```
?

tshemsedinov: But need to check, will it optimize stack or not

tshemsedinov: I mean optimize is shorten paths and remove line brakes and padding

SemenchenkoVitaliy: It will remove padding, but won't shorten paths: https://github.com/metarhia/metalog/blob/eb3f92c66f93ca5772563e047c7baded11770bb1/metalog.js#L58.

lib/db.mysql.js
```diff
@@ -51,7 +51,7 @@ if (api.mysql) {
     });
 
     connection.on('error', err => {
-      database.application.log.error(err);
+      database.application.log.error(err.toString());
```
SemenchenkoVitaliy:
```suggestion
      database.application.log.error(err.stack);
```
Such change should be performed everywhere across all files.

lib/db.mysql.js
```diff
@@ -27,7 +27,7 @@ if (api.mysql) {
 
     connection.connect(err => {
       if (err) {
-        database.application.log.error(err);
+        database.application.log.error(err.stack);
```
belochub: @tshemsedinov can we replace this and similar occurrences with this?
```js
database.application.logException(err)
```

belochub:
@JuliaGerasymenko there is one more place with a similar issue:
https://github.com/metarhia/impress/blob/4c5382a09446ef57422c1000f4cba22ed40ef681/lib/core.js#L524

belochub:
@tshemsedinov it is already being done in the `application.logException()`.

lib/core.js
```diff
@@ -521,7 +521,7 @@ impress.load = () => {
       impress.workerApplicationClient.access = { allowed: true };
       const file = impress.workerApplicationFile;
       impress.workerApplicationClient.runScript('worker', file, err => {
-        if (err) impress.log.error(err);
+        if (err) impress.log.error(err.stack);
```
tshemsedinov:
```suggestion
        if (err) impress.log.error(impress.shortenStack(err.stack));
```

tshemsedinov:
Landed in caac938

PR #1113, opened 2019-02-25
Fix race condition in tests

nechaido:


lib/core.js
```diff
@@ -747,3 +733,21 @@ impress.logApiMethod = fnPath => {
...
+      test.test(app.name, test => {
+        test.endAfterSubtests();
+        app.tests.unit
+          .concat(app.tests.integration)
+          .forEach(testCase => test.test(testCase.path, testCase.test));
+      });
+    });
+  });
+  api.test.runner.instance.on('finish', hasErrors => {
+    process.send({ name: 'impress:testsFinished', hasErrors });
```
SemenchenkoVitaliy: Shouldn't we wait for some time in case if there are tests that can fail after the `finish` event?

tshemsedinov:
Landed in 3f2098b

PR #1115, opened 2019-02-25
Add schema reloading

nechaido:


applications/example/schemas/version.js
```diff
@@ -0,0 +1,3 @@
+{
+  version: 1,
+}
```
tshemsedinov: It's ok for now but we need schema name, author, license, checksum in `version.js` and I'd rename it to `.schema` or something like `schema.pkg` @nechaido

tshemsedinov:
Need to rebase @nechaido

applications/example/schemas/version.js
lundibundi: Perhaps ms.js?

tshemsedinov:
Landed in 936b616

PR #1120, opened 2019-02-26
Fix `gs` initialization

nechaido:
@tshemsedinov PTAL.
AFAIK `database.connection` should be initialized with an actual connection to a db, but when I do so server stops working.

/cc @belochub


tshemsedinov:
Looks fine, but I need more investigation here to fix @nechaido

tshemsedinov:
Landed in d2c9194

PR #1121, opened 2019-02-26
Load globalstorage system schemas

nechaido:


lundibundi:
Why not add a `schemas.js` configuration file that will have `paths: []` of additional relative to application paths to load schemas from?

Also, does this work `api.path.join(api.gs.path, 'schemas', 'system')` or you planned to add `path` to the exports of `gs` later?

nechaido:
@lundibundi 
> Why not add a `schemas.js` configuration file that will have `paths: []` of additional relative to application paths to load schemas from?

Because gs schemas are located in node_modules, and its relative location may vary.

> Also, does this work `api.path.join(api.gs.path, 'schemas', 'system')` or you planned to add `path` to the exports of `gs` later?
It is already assigned by impress.

lundibundi:
> Because gs schemas are located in node_modules, and its relative location may vary.

I would say that `node_modules` location is pretty stable, I'm not sure of the conditions where it varies.

The other solution would be to specify modules/packets to load schemas from and then impress would `api.path.join(api[module].path, 'schemas')` and load them.

nechaido:
@lundibundi 
> I would say that `node_modules` location is pretty stable, I'm not sure of the conditions where it varies.

Relative path to executable may vary in a monorepo with dependencies installed by yarn or a usual repo with dependancies installed by npm.

lib/application.js
```diff
@@ -357,11 +358,18 @@ class Application extends api.events.EventEmitter {
         return;
       }
 
-      const { options, config } = getMsConfig();
+      const {
+        additionalSchemas: additionalPaths, options, config
+      } = getMsConfig();
+
+      const pathes = additionalPaths && additionalPaths.length ?
```
lundibundi:
```suggestion
      const paths = additionalPaths && additionalPaths.length ?
```

nechaido:
ping @belochub.

lib/application.js
```diff
@@ -357,11 +358,18 @@ class Application extends api.events.EventEmitter {
         return;
       }
 
-      const { options, config } = getMsConfig();
+      const {
+        additionalSchemas: additionalPaths, options, config
+      } = getMsConfig();
+
+      const paths = additionalPaths && additionalPaths.length ?
```
belochub:
```suggestion
      const paths = additionalPaths && additionalPaths.length !== 0 ?
```

tshemsedinov:
Landed in 8692113

PR #1122, opened 2019-02-28
Save latests schema

nechaido:


lib/application.js
```diff
@@ -378,11 +379,18 @@ class Application extends api.events.EventEmitter {
   //   callback <Function>
   //     err <Error> | <null>
   updateSchemas({ version }, callback) {
+    if (this.latestSchema && version < this.latestSchema.version) {
```
SemenchenkoVitaliy: I think there should also be check if `version === this.latestSchema.version` so it won't load the same schema again.

lib/application.js
```diff
@@ -23,6 +23,7 @@ class Application extends api.events.EventEmitter {
     this.config = new impress.Config(this);
     this.scheduler = new impress.Scheduler(this);
     this.schemas = new Map(); // Map of application schemas in metaschema format
+    this.latestSchema = null;
```
tshemsedinov: Why do we have both `latestSchema` and `schemas` fields?

nechaido: So that we can easily determine the latest available schema.

lib/application.js
```diff
@@ -378,11 +379,24 @@ class Application extends api.events.EventEmitter {
   //   callback <Function>
   //     err <Error> | <null>
   updateSchemas({ version }, callback) {
+    if (this.latestSchema) {
+      if (version === this.latestSchema.version) {
+        callback(null);
+        return;
+      }
+      if (version < this.latestSchema.version) {
+        application.log.warning('Attempt to update schema to a lower version');
```
lundibundi:
```suggestion
        application.log.warning('Attempt to update schema to a lower version. Ignored');
```
?

tshemsedinov:
Landed in e7f8f28

PR #1124, opened 2019-03-01
Restrict serving application folder index

primeare:
We need to consider whether user can see the static folder index. By now, he can.


tshemsedinov:
Landed in b4d3983

PR #1128, opened 2019-03-06
Implement hard drive smart error monitoring

bugagashenkj:


lib/health.js
```diff
@@ -170,6 +170,42 @@ const cmdExec = (cmd, callback) => {
   });
 };
 
+const getSmartStatusOfDrive = (driveName, callback) => {
+  const cmd = `smartctl -H /dev/${driveName} | grep overall`;
```
nechaido: I'd prefer to use `rg`, `ag` if either is installed.

@belochub WDYT?

lib/health.js
```diff
@@ -170,6 +170,42 @@ const cmdExec = (cmd, callback) => {
...
+  api.cp.exec(cmd, (error, stdout) => {
+    if (error) callback(error);
+
+    const status = stdout.toString().split(':').pop().trim();
+    if (status === 'PASSED') {
```
nechaido:
```suggestion
    callback(null, status === 'PASSED')
```

bugagashenkj:
@tshemsedinov Should we support disk monitoring for Windows?

lib/health.js
```diff
@@ -170,6 +170,38 @@ const cmdExec = (cmd, callback) => {
   });
 };
 
+const getSmartStatusOfDrive = (driveName, callback) => {
+  const cmd = `smartctl -H /dev/${driveName}`;
+  api.cp.exec(cmd, (error, stdout) => {
+    if (error) callback(error);
```
belochub: You don't return after calling the callback here, thus it is possible that the callback will be called twice here.
Also, I think it'd be better to not ignore `stderr`, and print it in this branch/pass it to the callback.

lib/health.js
```diff
@@ -170,6 +170,38 @@ const cmdExec = (cmd, callback) => {
...
+    callback(null, status === 'PASSED');
+  });
+};
+
+
+health.getSmartStatus = callback => {
+  const cmd = 'lsblk -d -J';
+  const drivesStatus = {};
+  api.cp.exec(cmd, (error, stdout) => {
+    if (error) callback(error);
```
belochub: Same thing here.

lib/health.js
```diff
@@ -170,6 +170,38 @@ const cmdExec = (cmd, callback) => {
...
+
+    const funcs = JSON.parse(stdout).blockdevices.map(drive =>
+      next => getSmartStatusOfDrive(drive.name, (error, status) => {
+        if (error) next(error);
```
belochub: Ditto.

lib/health.js
```diff
@@ -170,6 +170,53 @@ const cmdExec = (cmd, callback) => {
...
+const getSmartStatusOfDrive = (driveName, callback) => {
+  const cmd = `smartctl -H /dev/${driveName}`;
+  api.cp.exec(cmd, (error, stdout, stderr) => {
+    if (error) {
+      callback(error);
+      return;
+    }
+
+    if (stderr) {
+      callback(stderr);
```
belochub: `stderr` is a string here, thus you should either log it here or create an error based on it, not pass it to the callback

lib/health.js
```diff
@@ -170,6 +170,53 @@ const cmdExec = (cmd, callback) => {
...
+    if (error) {
+      impress.log.error(stderr);
+      callback(error);
+      return;
+    }
+
+    const funcs = JSON.parse(stdout).blockdevices.map(drive =>
+      next => getSmartStatusOfDrive(drive.name, (error, status) => {
+        if (error) {
+          impress.log.error(stderr);
```
belochub: Why do you log the `stderr` obtained when running `lsblk -d -J` when you get an error from `getSmarStatusOfDrive()`?
It is now possible for it to be logged more than once.

lib/health.js
```diff
@@ -170,6 +170,49 @@ const cmdExec = (cmd, callback) => {
...
+        if (error) {
+          next(error);
+          return;
+        }
+        drivesStatus[drive.name] = status;
+        next(null);
+      })
+    );
+
+    api.metasync.sequential(funcs, (error) => {
```
tshemsedinov:
```suggestion
    api.metasync.sequential(funcs, error => {
```

tshemsedinov:
Landed in 4050d2e

Issue #1108, opened 2019-02-20
Multiple calls of callback


SemenchenkoVitaliy:
In `api.db.pgsql.open` the callback is passed as an argument and it is called at the end of the function.

https://github.com/metarhia/impress/blob/4c5382a09446ef57422c1000f4cba22ed40ef681/lib/db.pgsql.js#L67

But if there was an error when trying to connect, `api.db.pgsql.open` will call recursively itself with the same arguments.

https://github.com/metarhia/impress/blob/4c5382a09446ef57422c1000f4cba22ed40ef681/lib/db.pgsql.js#L37-L47

In case if the connection failed `callback` will be called at least twice. Is this expected behaviour? I didn't happen to find usages of it.

The same goes for `oracle` and `mysql`.

Also, it seems that it will reset `database.retryCounter` to zero every time, so it will try to connect forever.

PR #1137, opened 2019-03-13
Fix database(mysql, pg, oracle) open scripts

lundibundi:
Fixes: https://github.com/metarhia/impress/issues/1108

### DIFF
--- a/lib/db.mysql.js
+++ b/lib/db.mysql.js
@@ -2,6 +2,30 @@
 
 // MySQL database connector for Impress Application Server
 
+const setupConnection = (database, connection, callback) => {
+  connection.on('query', (err, res, fields, query) => {
+    if (err) {
+      database.application.log.error(
+        `MySQL Error[${err.errno}]: ${err.code} ${query.sql}`
+      );
+    }
+    database.application.log.debug(query.sql);
+  });
+
+  connection.on('slow', (err, res, fields, query, executionTime) => {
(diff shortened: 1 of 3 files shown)


lib/db.mysql.js
```diff
@@ -31,36 +55,17 @@ if (api.mysql) {
...
           if (database.retryCounter <= database.retryCount) {
             api.db.mysql.open(database, callback);
+          } else {
+            callback(new Error('Cannot open connection'));
           }
         }, database.retryInterval);
+        return;
       }
       database.retryCounter = 0;
+      setupConnection(database, connection, callback);
```
SemenchenkoVitaliy: You still pass `callback` which will be called on connection error
https://github.com/metarhia/impress/blob/1a6197c5b9468d2782fcc80f9eaf39c572311cbf/lib/db.mysql.js#L19-L26
in addition to being called at the end of this function
https://github.com/metarhia/impress/blob/1a6197c5b9468d2782fcc80f9eaf39c572311cbf/lib/db.mysql.js#L67

lundibundi: Yes, that will lead to the re-establishment of a database connection after error, that's the current behavior, though all other DB connectors do not reconnect in case of error.
I can remove this if there is a consensus on this topic. Personally, I think it should align with other DB connectors (remove api.db.mysql.open from 'error' in mysql)

tshemsedinov:
Landed in a82d202

PR #1140, opened 2019-03-18
Fix impress.shutdown crash when no 'log' is present

lundibundi:


tshemsedinov:
Closed in favour of #1143

PR #1150, opened 2019-04-03
Fix usage of filename reserved by Windows

belochub:
Filename CON is reserved by Windows and using it leads to problems on
Windows systems.
See http://msdn.microsoft.com/en-us/library/aa365247.aspx for more
information.


tshemsedinov:
Landed in 8de11a6

PR #1151, opened 2019-04-03
Fix and refactor health.getDrives()

belochub:
Before it wasn't working correctly on Linux with locales other than
those that result in `df` outputting the column names in English.

Refs: https://github.com/metarhia/impress/issues/993


tshemsedinov:
Landed in 47ba39c

PR #1154, opened 2019-04-16
Simplify updateSchemas and allow any 'version' field

lundibundi:


lib/application.js
```diff
@@ -372,7 +372,6 @@ class Application extends api.events.EventEmitter {
       const loadSchema = api.util.callbackify(api.metaschema.fs.load);
       loadSchema(paths, options, config, (err, schema) => {
         if (err) {
-          this.logException(err);
```
tshemsedinov: Why don't log schema loading errors?

lundibundi: This function may be used by the user applications to manually load and update schemas, therefore this output will only clutter the logs while user can decide on the appropriate handling himself.

The error is now logged at https://github.com/metarhia/impress/pull/1154/files#diff-5372f626ee15242f1e2c6eb31655b4faR423 where we call this function.

nechaido:
@tshemsedinov ping.

lundibundi:
@tshemsedinov ping.

tshemsedinov:
Landed in https://github.com/metarhia/impress/commit/b1fa5a82c4943d62d66456e23c16681d070d7ba5

PR #1157, opened 2019-05-14
Implement application.broadcastToClients

lundibundi:
This will allow to easily send an event to all of the clients on each
worker of this application.

### DIFF
--- a/applications/example/test/jstp/utils.js
+++ b/applications/example/test/jstp/utils.js
@@ -1,12 +1,12 @@
 api.jstpTest = {
   host: '127.0.0.1',
-  port: 5000,
-  connect: cb => api.jstp.ws.connectAndInspect(
+  ports: [5000, 5001],
+  connect: (cb, port = api.jstpTest.ports[0]) => api.jstp.ws.connectAndInspect(
     'example',
     null,
     ['interfaceName'],
     {},
-    `ws://${api.jstpTest.host}:${api.jstpTest.port}`,
+    `ws://${api.jstpTest.host}:${port}`,
     cb
   ),
 };
--- a/lib/application.js
+++ b/lib/application.js
@@ -18,8 +18,6 @@ class Application extends api.events.EventEmitter {
     this.started = false;
     this.stopped = false;
     this.log = impress.logger.bind(name);
-    this.backend = api.common.emitter();
-    this.frontend = api.common.emitter();
     this.config = new impress.Config(this);
     this.scheduler = new impress.Scheduler(this);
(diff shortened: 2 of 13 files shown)


lib/application.js
```diff
@@ -57,6 +57,9 @@ class Application extends api.events.EventEmitter {
     const after = () => {
       if (--count === 0) started();
     };
+    this.frontend.on('broadcastToClients', (...args) => {
+      impress.server.instance.broadcast(...args);
```
belochub: Shouldn't there be a check that validates that `impress.server.instance` is a `jstp.Server` here?
Since `http.Server` doesn't have the `broadcast()` method, won't this fail with `TypeError: impress.server.instance.broadcast is not a function` when used with a worker serving HTTP?

lundibundi:
@tshemsedinov there seems to be quite a bit of changes for that, I think it'd be better to do that in a separate PR, wdyt?

applications/example/test/jstp/utils.js
```diff
@@ -1,12 +1,12 @@
 api.jstpTest = {
   host: '127.0.0.1',
-  port: 5000,
-  connect: cb => api.jstp.ws.connectAndInspect(
+  ports: [5000, 5001],
+  connect: (cb, port) => api.jstp.ws.connectAndInspect(
```
nechaido:
```suggestion
  connect: (cb, port = api.jstpTest.ports[0]) => api.jstp.ws.connectAndInspect(
```

lib/cloud.js
```diff
@@ -183,58 +179,36 @@ class Cloud extends api.events.EventEmitter {
...
-    });
-
     application.state.on('change', (data, isTarget) => {
       if (!isTarget) {
         api.cloud.sendState(application.name, data);
       }
     });
   }
 
-  sendEventToNode(eventName, data, nodeId) {
```
belochub: Removal of this method and the ones below breaks backward compatibility and has to be introduced in the next major version. Thus we should probably mark the PR as such, to avoid accidental publish of these changes without the major version increment.

lundibundi: Then I think we should also remove `frontend` and `backend` emitters completely?
/cc @tshemsedinov

tshemsedinov:
Landed in https://github.com/metarhia/impress/commit/fae083bbdfb7429360eb8c8636bff03737404f0b https://github.com/metarhia/impress/commit/a411f4f1a779ca49c4fc1a579815b0ad2765c0ef https://github.com/metarhia/impress/commit/221652251162da460757077703e1d10cf28cc29d

PR #1169, opened 2019-06-04
Allow custom script timeout in application config

lundibundi:
This allows to specify custom scriptTimeout in the application
config because currently hardcoded 500ms may not be enough to run
some scripts especially when running in constrained environment
(i.e. CI).


tshemsedinov:
Landed in https://github.com/metarhia/impress/commit/f98492f333d77bf61e01d8766038870517102042

PR #1170, opened 2019-06-06
Fix parent test timeout in test mode

lundibundi:
As of now metatests uses 30s as default test timeout therefore, because
impress in core.js when starting the subtests didn't change the timeout
value all tests were limited to being run no more that 30 seconds
in total and after this limit was reached would be forcefully finished.

This is obviously not enough for an extensive test suite.
This commit allows to specify a custom testTimeout in the application
config and disables timeout on the Impress parent test.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -43,6 +43,7 @@ impress.relative = path => path.substr(impress.dir.length);
 impress.DEFAULT_KEEP_ALIVE = api.common.duration('5s');
 impress.DEFAULT_TIMEOUT = api.common.duration('30s');
 impress.DEFAULT_SHUTDOWN = api.common.duration('5s');
+impress.APPLICATION_TEST_TIMEOUT = api.common.duration('5m');
 
 {
   const submodules = [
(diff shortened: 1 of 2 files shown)


lib/core.js
```diff
@@ -2,6 +2,8 @@
 
 // Impress Application Server Core
 
+const APPLICATION_TEST_TIMEOUT = 5 * 60 * 1000; // 5m
```
tshemsedinov: See below, all constants usually added to namespace `impress`, for example: `impress.DEFAULT_KEEP_ALIVE` so we need to move this constants to namespace or move all other constants to module scope.

lundibundi: Done.

tshemsedinov:
Landed in https://github.com/metarhia/impress/commit/b367ca5bb8bd4c9adc6494bff30970332c6b1e93

lib/core.js
```diff
@@ -753,6 +754,8 @@ impress.startTests = () => {
   api.test.test('Impress applications test', test => {
     test.endAfterSubtests();
     Object.values(impress.applications).forEach(app => {
+      const appConfig = app.config.sections.application;
+      const timeout = appConfig.testTimeout || impress.APPLICATION_TEST_TIMEOUT;
```
belochub: There is no possibility to specify 0 as a timeout value at the moment.

PR #1171, opened 2019-06-07
Version 1.0.8

tshemsedinov:
- Drop nodejs 8.x
- Support nodejs starting from 10.16.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,15 +47,15 @@
     "doc": "metadoc -c"
   },
   "engines": {
-    "node": ">=8.6.0"
+    "node": ">=10.16.0"
   },
   "dependencies": {
     "@metarhia/common": "^2.1.0",
     "@metarhia/jstp": "^2.3.0",
     "accept-language": "^3.0.18",
-    "argon2": "^0.24.0",
-    "concolor": "^0.1.12",
-    "csv-stringify": "^5.3.3",
+    "argon2": "^0.25.1",
+    "concolor": "^0.1.13",
+    "csv-stringify": "^5.3.4",
     "globalstorage": "^0.9.1",
     "iconv-lite": "^0.5.0",
     "mdsf": "^1.1.1",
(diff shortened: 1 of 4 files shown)


deploy/centos7x64-src.sh
```diff
@@ -7,15 +7,15 @@ yum -y install smartmontools
 yum -y install epel-release
 yum -y install certbot
 cd /usr/src
-wget https://nodejs.org/dist/v12.0.0/node-v12.0.0.tar.gz
-tar zxf node-v12.0.0.tar.gz
-rm -f ./node-v12.0.0.tar.gz
-cd node-v12.0.0
+wget https://nodejs.org/dist/v12.4.0/node-v12.4.0.tar.gz
```
lundibundi: I'd suggest to put a variable with `NODE_VERSION='v12.4.0'` at the top of such files and then use it as 
```bash
wget "https://nodejs.org/dist/${NODE_VERSION}/node-${NODE_VERSION}.tar.gz"
``` 
and so on to avoid changing all this boilerplate.

package.json
```diff
@@ -47,7 +47,7 @@
     "doc": "metadoc -c"
   },
   "engines": {
-    "node": ">=8.6.0"
+    "node": ">=10.16.0"
```
lundibundi: Why not 10.0.0 though?

tshemsedinov: `10.15.3` gives `Array.prototype.values()` support, returning `ArrayIterator`
`10.16.0` gives `await events.once(ee, 'eventName')` support

deploy/centos7x64-src.sh
tshemsedinov: But we will still change this variable in all files, maybe there are url to the latest ver?

tshemsedinov:
Maybe we can do this for other repos?

nechaido:
@tshemsedinov I take it
> Maybe we can do this for other repos?

Refers to

> 10.15.3 gives Array.prototype.values() support, returning ArrayIterator
10.16.0 gives await events.once(ee, 'eventName') support

IMHO, we should change it only once such a feature is needed in a repo.

tshemsedinov:
Landed in https://github.com/metarhia/impress/commit/20d52a73bbf447e717bd3d4f19671fae7b5fab3a
But this will not be published to npm
Version 1.0.9 will be published instead

PR #1173, opened 2019-06-14
Allow to disable timeout for application tests in application.js config

lundibundi:
This commit fixes the check of testTimeout to allow passing any value
as timeout. Therefore it is possible to put `testTimeout: 0` in
application.js config to disable the timeout for this application tests.

/cc @belochub


tshemsedinov:
Landed in https://github.com/metarhia/impress/commit/eee470dabcc009204fd0b11449628a0b1499243f

PR #1174, opened 2019-06-14
Prettify test output with TapReporter when possible

lundibundi:


tshemsedinov:
Not relevant for v2.x

PR #1175, opened 2019-06-14
WIP: Refactor install.js, separate example/test and limit publish artifacts

lundibundi:
Please look at the commit descriptions.

There are quite a few changes but most of them are from moving files in 
`Separate example and test applications f4efb98`, I'd suggest reviewing each commit separately.

In a nutshell `Separate example and test applications` copies, the `example` app into `test` and removes all example code from it while removing all unnecessary tests from the `example` app.

The last commit, therefore, removes all unnecessary artifacts from the published package. I tried installing it after `npm pack` and it worked just fine, though it may need some additional testing for installation via shell scripts and impress-cli.


lundibundi:
This will have to be remade from the ground up, no need to keep this open.

Issue #1176, opened 2019-06-20
`process.on('warning')` may happen earlier than `impress.log` will be initialized


o-rumiantsev:
https://github.com/metarhia/impress/blob/eee470dabcc009204fd0b11449628a0b1499243f/lib/core.js#L127-L130

For example, if `fs.promises` are used in project, there might be `warning` event emitted by `process` before `impress` will be started, and therefore logger won't be initialized. So there will be an error thrown: `TypeError: Cannot read property 'warn' of undefined`

tshemsedinov:
Need an example to reproduce this bug

lundibundi:
@tshemsedinov If you add a dependency that uses experimental api or in some other way causes a warning (i.e. add `require('fs').promises` in Node.js 10 to some dependency) it will crash impress started with `npm test` or `./server.sh` with
![Screenshot from 2019-06-25 21-20-24](https://user-images.githubusercontent.com/9109612/60123117-5d1fe680-978f-11e9-8bbb-729b57c77b65.png)

PR #1179, opened 2019-06-24
Fix `warning` event handling

o-rumiantsev:
Closes: https://github.com/metarhia/impress/issues/1176

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -126,7 +126,13 @@ process.on('uncaughtException', err => {
 
 process.on('warning', warning => {
   const stack = impress.shortenStack(warning.stack);
-  impress.log.warn(`${warning.message} ${stack}`);
+  const message = `${warning.message} ${stack}`;
+  if (impress.log) {
+    impress.log.warn(message);
+  } else {
+    // TODO: initialize logger before performing any actions that may require it
+    console.warn(message);
+  }
 });
 
 process.on('multipleResolves', (type, promise, repeated) => {


lib/core.js
```diff
@@ -126,7 +126,12 @@ process.on('uncaughtException', err => {
 
 process.on('warning', warning => {
   const stack = impress.shortenStack(warning.stack);
-  impress.log.warn(`${warning.message} ${stack}`);
+  const message = `${warning.message} ${stack}`;
+  if (impress.log) {
+    impress.log.warn(message);
+  } else {
+    console.warn(message);
```
lundibundi: Could you add a TODO here to note that this is just a fix to avoid a crash and not an actual solution to the problem. Which would be to move logger initialization to the very beginning of the impress and/or and some sort of buffer to it until it has fully initialized.

lib/core.js
```diff
@@ -126,7 +126,13 @@ process.on('uncaughtException', err => {
 
 process.on('warning', warning => {
   const stack = impress.shortenStack(warning.stack);
-  impress.log.warn(`${warning.message} ${stack}`);
+  const message = `${warning.message} ${stack}`;
+  if (impress.log) {
+    impress.log.warn(message);
+  } else {
+    // TODO: move logger initialization to the very beginning
```
nechaido:
```suggestion
    // TODO: initialize logger before performing any actions that may require it
```

nechaido:
@tshemsedinov release ASAP

tshemsedinov:
@nechaido :+1:

tshemsedinov:
Landed in https://github.com/metarhia/impress/commit/fe598aa9cdef4deb9e3ba502c7ab68d20bdb6e85

PR #1191, opened 2019-08-10
Fix impress instance shutting down on metalog reload

belochub:


tshemsedinov:
Landed in https://github.com/metarhia/impress/commit/5d713d08c6525efb11b9065ef6d95673b48de399

PR #1226, opened 2020-06-10
Named API interfaces with versioning

tshemsedinov:
Closes: https://github.com/metarhia/impress/issues/1215

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -88,11 +88,11 @@ class Application extends events.EventEmitter {
   }
 
   async loadDomain(fileName) {
-    const script = await this.createScript(fileName);
-    if (!script) return;
     const rel = fileName.substring(this.domainPath.length + 1);
     if (!rel.endsWith('.js')) return;
     const name = path.basename(rel, '.js');
+    const script = await this.createScript(fileName);
+    if (!script) return;
     const config = this.config.sections[name];
     const sandbox = this.domainSandbox;
     sandbox.application[name] = { config };
--- a/lib/application.js
+++ b/lib/application.js
@@ -67,7 +67,7 @@ class Application extends events.EventEmitter {
     const { sandbox } = session || this;
     const iface = this.api[iname];
     if (!iface) return null;
-    const version = ver === 'default' ? iface.default : parseInt(ver);
+    const version = ver === '*' ? iface.default : parseInt(ver);
     const methods = iface[version.toString()];
     if (!methods) return null;
     const script = methods[methodName];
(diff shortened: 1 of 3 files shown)


lib/application.js
```diff
@@ -80,36 +87,46 @@ class Application extends events.EventEmitter {
     }
   }
 
-  async loadScript(place, fileName) {
-    const { name, ext } = path.parse(fileName);
-    if (ext !== '.js' || name.startsWith('.')) return;
+  async loadDomain(fileName) {
+    const rel = fileName.substring(this.domainPath.length + 1);
+    const [name, ext] = rel.split('.');
```
lundibundi, line 92: This will fail for names with dots in it which is not that uncommon (i.e. `'name.with.dots.js'`).

lib/application.js
```diff
@@ -80,36 +87,46 @@ class Application extends events.EventEmitter {
...
-      sandbox.application[name] = { config };
-      const exp = script.runInContext(sandbox, SCRIPT_OPTIONS);
-      if (config) exp.config = config;
-      this.sandbox.application[name] = exp;
-      this.sandboxInject(name, exp);
-      if (exp.start) exp.start();
-    } else {
-      scripts.set(name, script);
-    }
+    if (!script) return;
```
lundibundi, line 95: Wouldn't it be better to throw/reject here (or in `createScript`) to have some sort of indication that something went wrong instead o silently finishing?

lib/application.js
```diff
@@ -61,9 +63,14 @@ class Application extends events.EventEmitter {
     }
   }
 
-  runScript(methodName, session) {
+  runScript(iname, ver, methodName, session) {
     const { sandbox } = session || this;
-    const script = this.api.get(methodName);
+    const iface = this.api[iname];
+    if (!iface) return null;
```
lundibundi, line 69: IMO It will be better to have this throw specific error codes to simplify debugging\logging of error cases.

lib/application.js
```diff
@@ -80,36 +87,46 @@ class Application extends events.EventEmitter {
...
+    if (config) exp.config = config;
+    sandbox.application[name] = exp;
+    this.sandboxInject(name, exp);
+    if (exp.start) exp.start();
+  }
+
+  async loadMethod(fileName) {
+    const rel = fileName.substring(this.apiPath.length + 1);
+    if (!rel.includes('/')) return;
+    const [[iname, ver], [name, ext]] = rel.split('/').map(s => s.split('.'));
```
lundibundi, line 109: Ditto

lib/application.js
```diff
@@ -80,36 +87,46 @@ class Application extends events.EventEmitter {
...
+    if (ext !== 'js') return;
+    const version = parseInt(ver, 10);
+    const script = await this.createScript(fileName);
+    if (!script) return;
```
lundibundi, line 113: Ditto

lib/application.js
tshemsedinov: We write logs in `Application.prototype.createScript` and return null, so if script is null we just do not need to load it

lib/application.js
tshemsedinov: If `Application.prototype.createScript` returns null we generate `not found` in `client.js`. I don't think that we need different error types like: interface not found, method not found.

lib/application.js
tshemsedinov: You are right, I'll rewrite this

tshemsedinov:
Landed in https://github.com/metarhia/impress/commit/9df1f354e38e799572e2a223830792dc50a0d397, https://github.com/metarhia/impress/commit/b2d6e2f4c9b3932476871713cbd985ac539371d9, https://github.com/metarhia/impress/commit/fd7951784158fb81484f3bf15190f710a05e20ca, https://github.com/metarhia/impress/commit/f61fffa97a83a9907a20a88c34fa022d34f0caf1, https://github.com/metarhia/impress/commit/b0d6acbabfd074da2450271482831de781241443 and https://github.com/metarhia/impress/commit/d15398fb808fb06cc8fd9b62b53bd1d1cadb23a0

PR #1264, opened 2020-07-26
Fix request consumption in server.js

lundibundi:
Previously stream errors during `req` consumption would lead to a hanged
connection and resource/memory leak.
This also removes redundant double-promise from receiveBody.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -18,14 +18,13 @@ const sample = arr => arr[Math.floor(Math.random() * arr.length)];
 
 const clients = new Map();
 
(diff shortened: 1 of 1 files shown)


lib/server.js
```diff
@@ -62,6 +61,9 @@ const listener = (req, res) => {
     }
     receiveBody(req).then(data => {
       client.message(data);
+    }, err => {
+      application.logger.error(err.stack);
+      client.error(500);
```
tshemsedinov, line 66: This will write logs: 
```suggestion
      client.error(500, err);
```

PR #1285, opened 2020-08-18
Fix config loader for submodules if no module file present

tshemsedinov:


tshemsedinov:
Landed in https://github.com/metarhia/impress/commit/534ed668020ffd32870a6bca554722e58c8b5219

Issue #1466, opened 2021-02-10
Add access to the environment variables from configs


imnetcat:
I'm frustrated when I can't set port in config like "process.env.PORT || 8000".

I would like to add support for heroku & may be other that using process.env.PORT

The best solution to this problem is to pass the port to the sandbox, which is then passed to the worker config.

imnetcat:
solution above gives us the opportunity write in the config like this
```
...
this.PORT || 8000
...
```

tshemsedinov:
In node.js starter kit we can just put such expressions to config: `({ port: process.env.PORT })` but after extracting it ti metavm we need to pass process to sandbox.

PR #1467, opened 2021-02-10
Add access to the environment variables from configs

imnetcat:
#1466 solution option
now we can write in the config like 
```
...
this.PORT || 8000
...
```

### DIFF
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -10,8 +10,7 @@ const { Server } = metarhia.metacom;
 const { Auth } = require('./auth.js');
 
 (async () => {
-  const sandbox = { process };
-  const context = metarhia.metavm.createContext(sandbox);
+  const context = metarhia.metavm.createContext({ process });
   const configPath = path.join(application.path, 'config');
   const options = { mode: process.env.MODE, context };
   const config = await new Config(configPath, options);
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -11,7 +11,7 @@ const { Auth } = require('./auth.js');
 
 (async () => {
   const sandbox = { process };
-  const context = metavm.createContext(sandbox);
+  const context = metarhia.metavm.createContext(sandbox);
   const configPath = path.join(application.path, 'config');
   const options = { mode: process.env.MODE, context };
   const config = await new Config(configPath, options);
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -10,12 +10,10 @@ const { Server } = metarhia.metacom;
 const { Auth } = require('./auth.js');
 
 (async () => {
-  const vm = require('vm');
-  const sandbox = { PORT: parseInt(process.env.PORT || 0) };
-  vm.createContext(sandbox);
-
+  const sandbox = { process };
+  const context = metavm.createContext(sandbox);
   const configPath = path.join(application.path, 'config');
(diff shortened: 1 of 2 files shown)


lib/worker.js
```diff
@@ -10,8 +10,12 @@ const { Server } = metarhia.metacom;
 const { Auth } = require('./auth.js');
 
 (async () => {
+  const vm = require('vm');
```
tshemsedinov, line 13:
```suggestion
```
All modules are loaded, see `dependencies.js`

lib/worker.js
```diff
@@ -10,8 +10,12 @@ const { Server } = metarhia.metacom;
...
+  const sandbox = { PORT: parseInt(process.env.PORT || 0) };
```
tshemsedinov, line 14:
```suggestion
  const sandbox = { process };
```
Not a good place for parsing ports, we can just pass process into sandbox here.

lib/worker.js
```diff
@@ -10,8 +10,12 @@ const { Server } = metarhia.metacom;
...
+  vm.createContext(sandbox);
+
```
tshemsedinov, line 16:
```suggestion
  const context = metavm.createContext(sandbox);
```
Use metavm here

lib/worker.js
```diff
@@ -10,8 +10,12 @@ const { Server } = metarhia.metacom;
...
   const configPath = path.join(application.path, 'config');
-  const options = { mode: process.env.MODE };
+  const options = { mode: process.env.MODE, sandbox };
```
tshemsedinov, line 18:
```suggestion
  const options = { mode: process.env.MODE, context };
```
We will rename sandbox to context in `metarhia/config` first, so `sandbox` will be used just in `impress` or other upper-level code but both `metavm` and `metarhia/config` will use `{ context }` in `options`

tshemsedinov (changes requested):
We will change it in a different way

imnetcat:
Okey, great

tshemsedinov:
Will be landed after this fix https://github.com/metarhia/config/issues/67

tshemsedinov:
To be landed after https://github.com/metarhia/impress/pull/1468

lib/worker.js
```diff
@@ -10,8 +10,10 @@ const { Server } = metarhia.metacom;
 const { Auth } = require('./auth.js');
 
 (async () => {
+  const sandbox = { process };
+  const context = metarhia.metavm.createContext(sandbox);
```
tshemsedinov, line 14:
```suggestion
  const context = metarhia.metavm.createContext({ process });
```

lib/worker.js
```diff
@@ -10,8 +10,9 @@ const { Server } = metarhia.metacom;
 const { Auth } = require('./auth.js');
 
 (async () => {
+  const context = metarhia.metavm.createContext({ process });
   const configPath = path.join(application.path, 'config');
```
tshemsedinov, line 14:
```suggestion
  const configPath = path.join(application.path, 'config');
  const context = metarhia.metavm.createContext({ process });
```

tshemsedinov:
Landed in https://github.com/metarhia/impress/commit/4c4098802de80be220bf955eab64e9492679456c

Issue #1761, opened 2022-07-20
Run plugins in sandboxes


georgolden:
### Impress and Node.js versions

node 16, impress 3.0.0-alpha.2

### Platform

Linux

### Describe the bug

There is no `context` in generated CRUD and there is no way to provide a context to generated CRUD rpc method. I tried to implement reach-featured CRUD plugin for impress, but I am unable to work with `context` and extract specific information about accountId and roles, cause there are no context inside those endpoints.

We need to reimplement plugin handling and for each async fucntion create a separate Procedure.
https://github.com/metarhia/impress/blob/c21d862f640d46a6ff4d9e28f7f61526db2b09c0/lib/interfaces.js#L68-L89

We are creating an interface from plugin config with a diff from normal interface. I think the problem is in the Procedure implementation or loading plugin.

georgolden:
https://github.com/georgolden/testPlugin

PR #1762, opened 2022-07-20
Load plugins via application sandbox

georgolden:
This RP is a start for bugfix https://github.com/metarhia/impress/issues/1761

Here is bug reproduced

### DIFF
--- a/lib/dependencies.js
+++ b/lib/dependencies.js
@@ -34,6 +34,12 @@ const assignNamespace = (container, name, value) => {
   container[camelKey] = value;
 };
 
+const loadPlugins = (lib) => {
+  for (const plugin of Object.keys(lib.plugins)) {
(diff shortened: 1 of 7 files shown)


georgolden:
@tshemsedinov there are no dev dependencies can be loaded by `impress/lib/dependencies.js`, so I decided to make a fix in a separate commit. 03521edeb953b588b98b425a52ad5ef97d3d9f1b . It can be cherry picked, rest commits are only for testing and there is no easy way to mock plugin, so I decided to have a temporary dependency for me to be able to solve an issue quickly.

If we can destruct plugin loading from the normal API methods loading, we can achieve nicer, testable version as well.

tshemsedinov (approved):

tshemsedinov:
Landed in https://github.com/metarhia/impress/commit/a9119797d50fcf71c69e20edae9b2a1435e9951f

PR #1828, opened 2023-05-06
Add Content-Length header for servces.js

Slavakaderkin:

### DIFF
--- a/lib/services.js
+++ b/lib/services.js
@@ -8,7 +8,8 @@ const { Schema } = require('metaschema');
 const request = (url, { method, body }) =>
   new Promise((resolve, reject) => {
     const proto = url.startsWith('https') ? https : http;
-    const headers = { 'Content-Type': 'application/json', 'Content-Length': body ? Buffer.byteLength(body) : 0 };
+    const len = body ? Buffer.byteLength(body) : 0;
+    const headers = { 'Content-Type': 'application/json', 'Content-Length': len };
     const req = proto.request(url, { method, headers }, (res) => {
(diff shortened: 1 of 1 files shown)


lib/services.js
```diff
@@ -8,7 +8,7 @@ const { Schema } = require('metaschema');
 const request = (url, { method, body }) =>
   new Promise((resolve, reject) => {
     const proto = url.startsWith('https') ? https : http;
-    const headers = { 'Content-Type': 'application/json' };
+    const headers = { 'Content-Type': 'application/json', 'Content-Length': body ? Buffer.byteLength(body) : 0 };
```
tshemsedinov, line 11:
```suggestion
    const len = body ? Buffer.byteLength(body) : 0;
    const headers = { 'Content-Type': 'application/json', 'Content-Length': len };
```

lib/services.js
```diff
@@ -8,7 +8,8 @@ const { Schema } = require('metaschema');
 const request = (url, { method, body }) =>
   new Promise((resolve, reject) => {
     const proto = url.startsWith('https') ? https : http;
-    const headers = { 'Content-Type': 'application/json' };
+    const len = body ? Buffer.byteLength(body) : 0;
+    const headers = { 'Content-Type': 'application/json', 'Content-Length': len };
```
tshemsedinov, line 12:
```suggestion
    const mimeType = 'application/json';
    const len = body ? Buffer.byteLength(body) : 0;
    const headers = { 'Content-Type': mimeType, 'Content-Length': len };
```

Issue #1707, opened 2022-01-14
Unable to require different version of dependency that impress depens on


lundibundi:
**Describe the bug**
If `impress` has dependency on a package it is impossible to install a different version of that package for the application.

**To Reproduce**
Impress has dependency on metatests https://github.com/metarhia/impress/blob/8bfd071831288526aa72b79e298f210ff0e3361a/package.json#L85
If one need to install newer (or older) version it will not work.
```
npm install metatests@0.6.6

// start impress application
// npm.metatests is still 0.7.2
```

**Expected behavior**
It is possible to use any version of package for application.

**Screenshots**

**Desktop (please complete the following information):**

- OS: Linux zou 5.15.13-arch1-1 SMP PREEMPT Wed, 05 Jan 2022 16:20:59 +0000 x86_64 GNU/Linux
- Node.js version 14.18.2
- Impress version 2.6.6

PR #1881, opened 2023-07-31
Require modules from application, not from impress

tshemsedinov:
Closes: https://github.com/metarhia/impress/issues/1707

- [ ] update .d.ts typings

### DIFF
--- a/lib/deps.js
+++ b/lib/deps.js
@@ -1,7 +1,11 @@
 'use strict';
 
+const CWD = process.cwd();
+
 const wt = require('node:worker_threads');
+const { createRequire } = require('node:module');
 const metautil = require('metautil');
+const appRequire = createRequire(`file://${CWD}/server.js`);
 
 const node = {};
 const npm = {};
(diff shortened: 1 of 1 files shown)


lib/deps.js
```diff
@@ -1,7 +1,11 @@
 'use strict';
 
+const CWD = process.cwd();
+
 const wt = require('node:worker_threads');
+const { createRequire } = require('node:module');
 const metautil = require('metautil');
+const appRequire = createRequire(`file://${CWD}/server.js`);
```
lundibundi, line 8: What if the user starts app from some nested folder? Do we perhaps want to try "search up the tree" to find package.json?

lib/deps.js
```diff
@@ -1,7 +1,11 @@
 'use strict';
 
+const CWD = process.cwd();
```
lundibundi, line 3: Do we still not support TS application? as this will probably not work with `dist`.

lib/deps.js
tshemsedinov: This solution expected  all metarhia projects to be of the same structure and they should have server.js in app root folder and I am ok with this

lib/deps.js
tshemsedinov: We don't support TS but TS may have different compile destination. We expect all app code separated to .js + .d.ts. TypeScript can check js files against types.

tshemsedinov:
Landed in https://github.com/metarhia/impress/commit/e0e016e1417ee269704d99df48088bc1ebe38d88

Issue #1740, opened 2022-05-20
Don't create new MessageChannel


tshemsedinov:
## Is your feature request related to a problem? Please describe.

Adding tasks in scheduler or invoke procedures from thread pool we use `new MessageChannel` to return result back. It's obvious overhead, so we already have open `MessagePort` to communicate.

## Describe the solution you'd like

Reuse `MessagePort` and rewrite message routing for that.

PR #1886, opened 2023-08-03
Fix MessagePort leak

nechaido:
Generally, it would be better to open only one MessagePort pair between main and worker threads. But this will fix a leak for now/ 

- [ ] update .d.ts typings

Here is the summery of the memory snapshots of a dev server after a few hours of running a scheduled task every 10s.

Without the fix:
Screenshot 2023-08-02 at 16 46 24

With the fix:
Screenshot 2023-08-02 at 16 46 36


tshemsedinov:
Related issue: https://github.com/metarhia/impress/issues/1740

Issue #1890, opened 2023-08-30
Restart on EADDRINUSE


tshemsedinov:
### Describe the problem

```
uncaughtException: Error: listen EADDRINUSE: address already in use 0.0.0.0:8000
  Server.setupListenHandle [as _listen2] (node:net:1463:16)
```

### Describe the solution

Port is in use by another server, so we can:
- wait for timeout
- try 3 times
- exit 1

MarhiievHE:
@tshemsedinov  It looks like the HTTP server is not returning this error via the error listener: https://github.com/metarhia/metacom/blob/c8565a96eb0b77a536b15f643f4a727321478972/lib/server.js#L213. Will it be normal to catch this error in this way in worker.js:

```js
    process.on('uncaughtException', function(err) {
      if (err.code === 'EADDRINUSE') {
        console.warn(`Address in use: ${options.host}:${port}, retry...`);
        const timer = setTimeout(() => {
          console.log({tryToStartCounter});
          tryToStartCounter++;
          if (tryToStartCounter > 3) process.kill(process.pid, 'SIGINT');
          application.server = new metarhia.metacom.Server(application, options);
        },options.timeouts.bind);
      }
    });
```

MarhiievHE:
@tshemsedinov It looks like the HTTP server is not returning this error via the error listener. Will it be ok to catch this error in this way in worker.js, or do we need to find a more elegant solution?:
```js
    process.on('uncaughtException', function(err) {
      if (err.code === 'EADDRINUSE') {
        console.warn(`Address in use: ${options.host}:${port}, retry...`);
        const timer = setTimeout(() => {
          console.log({tryToStartCounter});
          tryToStartCounter++;
          if (tryToStartCounter > 3) process.kill(process.pid, 'SIGINT');
          application.server = new metarhia.metacom.Server(application, options);
        },options.timeouts.bind);
      }
    });
```

MarhiievHE:
> @tshemsedinov It looks like the HTTP server is not returning this error via the error listener. Will it be ok to catch this error in this way in worker.js, or do we need to find a more elegant solution?:
> 
> ```js
>     process.on('uncaughtException', function(err) {
>       if (err.code === 'EADDRINUSE') {
>         console.warn(`Address in use: ${options.host}:${port}, retry...`);
>         setTimeout(() => {
>           tryToStartCounter++;
>           if (tryToStartCounter > 3) process.kill(process.pid, 'SIGINT');
[...]

PR #1907, opened 2023-09-29
Fix restart on EADDRINUSE

MarhiievHE:
Suggestions for solving the problem: https://github.com/metarhia/impress/issues/1890

### DIFF
--- a/impress.js
+++ b/impress.js
@@ -183,7 +183,7 @@ const stopApplication = (dir) => {
   }
 };
 
-async function stop(code = 0) {
+const stop = async (code = 0) => {
   const portsClosed = new Promise((resolve) => {
     impress.console.info('Graceful shutdown in worker 0');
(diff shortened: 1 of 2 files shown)


impress.js
```diff
@@ -178,7 +183,7 @@ const stopApplication = (dir) => {
   }
 };
 
-const stop = async () => {
+async function stop(code = 0) {
```
tshemsedinov, line 186:
```suggestion
const stop = async (code = 0) => {
```

MarhiievHE: But if we leave the arrow function, then the code will not pass the test. I actually have a question: should we terminate the impress at once? We can have other applications and if one of them has a problem, then in this case we terminate all applications.

tshemsedinov:
Closes in favour of https://github.com/metarhia/impress/pull/1908 and https://github.com/metarhia/metacom/pull/459

Issue #1938, opened 2023-11-11
Cannot set property x of #<Object> which has only a getter


DmitryScaletta:
### Impress and Node.js versions

impress: 3.0.13, node: v21.1.0

### Describe the bug

[drizzle-orm@0.29.0](https://github.com/drizzle-team/drizzle-orm/releases/tag/0.29.0)

First the module loader loads the main package which imports a `relations` module which exports a `relations` function.

<details>
<summary>node_modules/drizzle-orm/index.cjs</summary>

https://github.com/drizzle-team/drizzle-orm/blob/0d833d1ca78735e29b124e2a672bda2c85738f48/drizzle-orm/src/index.ts#L10

![image](https://github.com/metarhia/impress/assets/5096735/7eeebe18-c16c-4817-af2d-8cdb59d3f73a)
</details>

<details>
<summary>node_modules/drizzle-orm/relations.cjs</summary>

https://github.com/drizzle-team/drizzle-orm/blob/0d833d1ca78735e29b124e2a672bda2c85738f48/drizzle-orm/src/relations.ts#L495

![image](https://github.com/metarhia/impress/assets/5096735/3dbbdb1b-75b0-4dbe-9c90-12b7fb5c7dd3)
</details>

Then it tries to load a `relations` submodule (the same name as the function that was already loaded).

<details>
<summary>node_modules/drizzle-orm/package.json</summary>

![image](https://github.com/metarhia/impress/assets/5096735/79cee473-d124-43cc-ac52-6c934f6fa543)
</details>

So `lib[subName]` is already exists and it points to the same object as `sub.relations`.

I added this debug code here:

https://github.com/metarhia/impress/blob/4cabe81e1cef6870a10d7c7b020f8618db3bdbf3/lib/deps.js#L57-L58

```js
const sub = appRequire(name + '/' + subName);
if (name === 'drizzle-orm' && subName === 'relations') {
  console.log(lib[subName] && (lib[subName] === sub[subName])); // true
}
lib[subName] = sub; // the error is here
```

```
TypeError: Cannot set property relations of #<Object> which has only a getter
    at loadModule (node_modules\impress\lib\deps.js:63:24)
    at Object.<anonymous> (node_modules\impress\lib\deps.js:89:13)
    at Module._compile (node:internal/modules/cjs/loader:1376:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)
    at Module.load (node:internal/modules/cjs/loader:1207:32)
[...]

Issue #1937, opened 2023-11-10
Module loader is trying to load submodules that use optional dependencies


DmitryScaletta:
### Describe the bug

By default the module loader loads all submodules in an `exports` field of a `package.json` file.
But some packages can have optional peer dependencies, and some of it submodules can use different optional dependencies.

For example [drizzle-orm](https://github.com/drizzle-team/drizzle-orm).

<details>
<summary>package.json</summary>

```json
{
  "peerDependencies": {
    "@aws-sdk/client-rds-data": ">=3",
    "@cloudflare/workers-types": ">=3",
    "@libsql/client": "*",
    "@neondatabase/serverless": ">=0.1",
    "@opentelemetry/api": "^1.4.1",
    "@planetscale/database": ">=1",
    "@types/better-sqlite3": "*",
    "@types/pg": "*",
    "@types/sql.js": "*",
    "@vercel/postgres": "*",
    "better-sqlite3": ">=7",
    "bun-types": "*",
    "knex": "*",
    "kysely": "*",
    "mysql2": ">=2",
    "pg": ">=8",
    "postgres": ">=3",
    "sql.js": ">=1",
    "sqlite3": ">=5"
  },
  "peerDependenciesMeta": {
    "mysql2": {
      "optional": true
    },
    "@vercel/postgres": {
      "optional": true
    },
    "better-sqlite3": {
      "optional": true
    },
    "@types/better-sqlite3": {
      "optional": true
    },
    "sqlite3": {
      "optional": true
    },
    "sql.js": {
      "optional": true
    },
    "@types/sql.js": {
      "optional": true
    },
    "@cloudflare/workers-types": {
      "optional": true
    },
    "pg": {
      "optional": true
    },
    "@types/pg": {
      "optional": true
    },
    "postgres": {
      "optional": true
    },
    "@neondatabase/serverless": {
      "optional": true
    },
    "bun-types": {
      "optional": true
    },
    "@aws-sdk/client-rds-data": {
      "optional": true
    },
    "@planetscale/database": {
      "optional": true
    },
    "knex": {
      "optional": true
    },
    "kysely": {
      "optional": true
    },
    "@libsql/client": {
      "optional": true
    },
    "@opentelemetry/api": {
      "optional": true
    }
  },
}
```
</details>

For postgres you only need submodules `drizzle-orm/node-postgres` and `drizzle-orm/pg-core` that require the `pg` package.

[...]

PR #1940, opened 2023-11-12
Improve loading of submodules

DmitryScaletta:
### 1. Don't load submodules with not installed optional dependencies

Some packages may have optional dependencies.
If there was an error during loading a submodue, and it's a missing optional dependency error, just skip this submodule.

Closes https://github.com/metarhia/impress/issues/1937

### 2. Skip submodules if they have already been loaded as a part of main module

Example: `some-module`

`package.json`

```js
{
  "exports": {
    ".": {
      // ...
      "default": "./index.js"
    },
    "./submodule": {
      // ...
      "default": "./submodule.js"
    }
  }
}
```

`index.js`

```js
export * from './submodule';
```

`submodule.js`

```js
export const submodule = () => {};
```

The main module exports all from the submodule.
And the submodule exports a function with the same name as a submodule name.
So when loading the submodule, `lib` already have a property `submodule`.

Closes https://github.com/metarhia/impress/issues/1938

- [ ] update .d.ts typings

### DIFF
--- a/lib/deps.js
+++ b/lib/deps.js
@@ -54,8 +54,16 @@ const loadModule = (name) => {
   const subKeys = Object.keys(pkg.exports).map((key) => key.substring(2));
   const subNames = subKeys.filter(validSubmodules);
   for (const subName of subNames) {
-    const sub = appRequire(name + '/' + subName);
-    lib[subName] = sub;
+    try {
+      const sub = appRequire(name + '/' + subName);
+      lib[subName] = sub;
+    } catch (e) {
+      if (e.message.startsWith("Cannot find module '")) {
+        const moduleName = e.message.substring(20, e.message.indexOf("'\n"));
+        const optional = pkg.peerDependenciesMeta?.[moduleName].optional;
+        if (optional) continue; else throw e;
+      }
+    }
   }
   return lib;
 };
--- a/lib/deps.js
+++ b/lib/deps.js
@@ -56,12 +56,14 @@ const loadModule = (name) => {
   for (const subName of subNames) {
     try {
       const sub = appRequire(name + '/' + subName);
+      if (lib[subName] && lib[subName] === sub[subName]) continue;
       lib[subName] = sub;
     } catch (e) {
       if (e.message.startsWith("Cannot find module '")) {
         const moduleName = e.message.substring(20, e.message.indexOf("'\n"));
         const optional = pkg.peerDependenciesMeta?.[moduleName].optional;
-        if (optional) continue; else throw e;
+        if (optional) continue;
+        else throw e;
       }
     }
   }
--- a/lib/deps.js
+++ b/lib/deps.js
@@ -61,7 +61,7 @@ const loadModule = (name) => {
     } catch (e) {
       if (e.message.startsWith("Cannot find module '")) {
         const moduleName = e.message.substring(20, e.message.indexOf("'\n"));
-        const optional = pkg.peerDependenciesMeta?.[moduleName].optional;
+        const optional = pkg.peerDependenciesMeta?.[moduleName]?.optional;
         if (optional) continue;
         else throw e;
       }


lib/deps.js
```diff
@@ -54,8 +54,18 @@ const loadModule = (name) => {
   const subKeys = Object.keys(pkg.exports).map((key) => key.substring(2));
   const subNames = subKeys.filter(validSubmodules);
   for (const subName of subNames) {
-    const sub = appRequire(name + '/' + subName);
-    lib[subName] = sub;
+    try {
+      const sub = appRequire(name + '/' + subName);
+      if (lib[subName] && lib[subName] === sub[subName]) continue;
```
tshemsedinov, line 59: Duplicate expression: `lib[subName]`

lib/deps.js
```diff
@@ -54,8 +54,18 @@ const loadModule = (name) => {
...
+      lib[subName] = sub;
+    } catch (e) {
+      if (e.message.startsWith("Cannot find module '")) {
```
tshemsedinov, line 62: - We do not use double quotas `"`, use linter before commit
- Parsing error in a such way is not reliable, it's better to get text between two double quotas with `metautil.between(e.message, '"', '"')` see: https://github.com/metarhia/metautil/tree/master#strings-utilities
- `e` is a bad naming, better `err` or even `error`

lib/deps.js
```diff
@@ -54,8 +54,18 @@ const loadModule = (name) => {
...
+        const moduleName = e.message.substring(20, e.message.indexOf("'\n"));
```
tshemsedinov, line 63: 20 is a magic number

lib/deps.js
```diff
@@ -54,8 +54,18 @@ const loadModule = (name) => {
...
+        const optional = pkg.peerDependenciesMeta?.[moduleName]?.optional;
```
tshemsedinov, line 64: Double optional chaining is not ok in out codebase

lib/deps.js
```diff
@@ -54,8 +54,18 @@ const loadModule = (name) => {
...
+        if (optional) continue;
+        else throw e;
```
tshemsedinov, line 66: You can just remove `else` here

tshemsedinov (changes requested):

lib/deps.js
DmitryScaletta: This check is actually necessary here because in every other case (except the one I described above) `lib[subName]` and `sub[subName]` would be equal to `undefined`.

lib/deps.js
DmitryScaletta: I used linter before commit.
It didn't fix the quotes.
Looks like prettier wants to use double quotes if string contains a single quote character: `'`
image

tshemsedinov:
Going to land this @DmitryScaletta but we need unittests. I'll propose unittests ASAP

lib/deps.js
```diff
@@ -54,8 +54,18 @@ const loadModule = (name) => {
...
+      if (lib[subName] && lib[subName] === sub[subName]) continue;
+      lib[subName] = sub;
+    } catch (err) {
+      if (err.code === 'MODULE_NOT_FOUND' && pkg.peerDependenciesMeta) {
+        const moduleName = metautil.between(err.message, `'`, `'`);
+        const optional = pkg.peerDependenciesMeta[moduleName]?.optional;
+        if (optional) continue;
+      }
+      throw err;
+    }
```
tshemsedinov, line 69:
```suggestion
    try {
      const sub = appRequire(`${name}/${subName}`);
      const target = lib[subName];
      if (target && target === sub[subName]) continue;
      lib[subName] = sub;
    } catch (error) {
      if (error.code === 'MODULE_NOT_FOUND' && pkg.peerDependenciesMeta) {
        const moduleName = metautil.between(error.message, `'`, `'`);
        const meta = pkg.peerDependenciesMeta[moduleName];
        if (!meta || meta.optional) continue;
      }
      throw error;
    }
```

Issue #1945, opened 2023-12-17
Local API endpoint queue settings doesn't work


KLarpen:
### Impress and Node.js versions

Impress: 3.0.13 Node: 20.9.0, 18.18.2

### Platform

macOS 14.1.2 Darwin Kernel Version 23.1.0: Mon Oct  9 21:27:24 PDT 2023; root:xnu-10002.41.9~6/RELEASE_ARM64_T6000 arm64

### Describe the bug

Local settings of the queue, placed directly into API endpoint module, not working at all. Any values there just doesn't applied, so the queue restrictions on the endpoint doesn't applied as well.

### To Reproduce

1. Start with [Example project](https://github.com/metarhia/Example/tree/master)
2. Check that general queue config at `application/config/server.js` contains

https://github.com/metarhia/impress/blob/4cabe81e1cef6870a10d7c7b020f8618db3bdbf3/test/config/server.js#L15-L19

3. Put file `application/api/example/testQueue.js` 

```js
({
  access: 'public',
  queue: {
    concurrency: 1,
    size: 1,
    timeout: 1500,
  },
  async method() {
    await node.timers.promises.setTimeout(2900);
    return { returned: true };
  },
});
```

4. Run server
5. Open browser tab with DevTools (console & network tabs in it)
6. Check that WS request `api` is active and there is messages 
7. Type in console `await api.example.testQueue();` and call it 3 times quickly (without awaiting for results)
8. All of request will return successfully after obvious concurrent execution despite our local queue setup.
```json
{"type":"call","id":4,"method":"example/testQueue","args":{}} 
{"type":"call","id":5,"method":"example/testQueue","args":{}}
{"type":"call","id":6,"method":"example/testQueue","args":{}}
{"type":"callback","id":4,"result":{"returned":true}}
{"type":"callback","id":5,"result":{"returned":true}}
{"type":"callback","id":6,"result":{"returned":true}}
``` 

### Expected behavior

Local queue settings of the API endpoint must be applied to control behaviour. Reproducing example above we expect :
- first return from the last request due to full queue without possibility to add any more
- second return from the 2nd request due to reached timeout awaiting in the queue
- 3rd return from the 1st request with successfull result of the long running function.

### Additional context

I'm already investigating possible fix to make pull request.

PR #1898, opened 2023-09-08
Move semaphore from metacom to `Application`

tshemsedinov:
Refs: https://github.com/metarhia/metacom/issues/449

- [x] tests and linter show no problems (`npm t`)
- [ ] tests are added/updated for bug fixes and new features
- [x] code is properly formatted (`npm run fmt`)
- [ ] description of changes is added in CHANGELOG.md
- [ ] update .d.ts typings

PR #1569, opened 2021-05-23
Custom API modules

tshemsedinov:
- [x] tests and linter show no problems (`npm t`)
- [x] code is properly formatted (`npm run fmt`)
- [x] description of changes is added in CHANGELOG.md
- [x] update .d.ts typings

PR #1946, opened 2023-12-17
Fix Procedure level queue

KLarpen:
Closes: https://github.com/metarhia/impress/issues/1945

API endpoint local queue params not working at all. I had investigate that the problem in Procedure class and originated from commit  "Improve Procedure class PR-URL: https://github.com/metarhia/impress/pull/1569 " https://github.com/metarhia/impress/pull/1569/commits/434e6c111fd2dd9da04b6e2e5605cafd07098b1d Where was applied changes into constructor:
- `Object.assign(this, exp);` removed
- expected property `concurrency` of the API endpoint was replaced with `queue` object containing `queue.concurrency`
- BUT `this.concurrency` not assigned directly anymore (as for example done for `this.parameters`.

However methods `enter` and `leave` still checking`this.concurrency` for existance as of today. 
https://github.com/metarhia/impress/blob/4cabe81e1cef6870a10d7c7b020f8618db3bdbf3/lib/procedure.js#L42-L57

As a result local Semaphore of the Procedure never applied. There should be check for the property `this.semaphore` instead. It suitable for this purpose because it will be `null` in case API endpoint doesn't provide queue settings.   

Before applying the fix I also added test case `lib/procedure queue` in [/test/procedure.js](https://github.com/metarhia/impress/blob/master/test/procedure.js) . By the way I had fixed application stub in Procedure tests to make it comply with updated contract (was changed by https://github.com/metarhia/impress/pull/1898).

- [x] tests are added/updated for bug fixes and new features
- [x] description of changes is added in CHANGELOG.md
- [ ] update .d.ts typings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased][unreleased]
 
+- Fixed API endpoints local queue settings applying
+
 ## [3.0.13][] - 2023-10-22
 
 - Fix serve static not in cache (e.g. certbot challenge)
--- a/lib/procedure.js
+++ b/lib/procedure.js
@@ -41,7 +41,7 @@ class Procedure {
 
   async enter() {
     await this.application.semaphore.enter();
-    if (this.concurrency) {
+    if (this.semaphore) {
       try {
         await this.semaphore.enter();
       } catch (error) {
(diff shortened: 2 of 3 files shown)


KLarpen:
Please review @tshemsedinov . This fix didn't change contracts and looks like a good candidate for next patch release of Impress.

tshemsedinov (approved):
LGTM

Issue #1947, opened 2023-12-18
Global config.timeouts.request setting doesn't applied


KLarpen:
### Impress and Node.js versions

Impress: 3.0.13, Metautil ^3.15.0, Node: 20.9.0

### Platform

macOS 14.1.2 Darwin Kernel Version 23.1.0: Mon Oct 9 21:27:24 PDT 2023; root:xnu-10002.41.9~6/RELEASE_ARM64_T6000 arm64

### Describe the bug

Configuration file `config/server.js` expected from an application folder contains property `timeouts.request`
https://github.com/metarhia/impress/blob/4cabe81e1cef6870a10d7c7b020f8618db3bdbf3/test/config/server.js#L7-L11

The documentation describes its purpose as ["API timeout"](https://github.com/metarhia/Docs/blob/dcbc596e43ae546ea9b136f2a68c72ae447e7513/content/en/START.md?plain=1#L92). However the value of config.timeouts.request actually doesn't apply that behavior. The problem has been initially discussed at [community call 133](https://youtu.be/2sWbuuuU_MY?t=4548). 

### To Reproduce

1. Start with [Example project](https://github.com/metarhia/Example/tree/master)
2. Check that general `timeouts` config at `application/config/server.js` contains `request: 5000`
3. Put file `application/api/example/testTimeout.js` 

```js
({
  access: 'public',
  async method() {
    await node.timers.promises.setTimeout(6000);
    return { returned: true };
  },
});
```

4. Run server
5. Open browser tab with DevTools (console & network tabs in it)
6. Check that WS request `api` is active and there is messages 
7. Type in console `await api.example.testTimeout();` and run
8. You will receive successful answer after 6 seconds

```json
{"type":"call","id":4,"method":"example/testTimeout","args":{}}
{"type":"callback","id":4,"result":{"returned":true}}
```
The server log contains nothing special: just log of the successful request to `example/testTimeout`.

9. Try to rise execution time in `application/api/example/testTimeout.js` by changing to `setTimeout(8000)`. Save and check server log for applied endpoint reload.
10. Call again `await api.example.testTimeout();` 
11. After 7 seconds browser console will show you `metacom.js:181 Uncaught Error: Request timeout` that [originated from frontend metacom client](https://github.com/metarhia/Example/blob/2435b8e643af3fc4a62fea1f3f20bdedc1e1d9c1/application/static/metacom.js#L4)
12. However after 8 seconds from request you still receive successful answer by the server

```json
[...]

PR #1948, opened 2023-12-19
WIP: Reimplement global timeouts.request usage

KLarpen:
Closes: https://github.com/metarhia/impress/issues/1947

- [x] tests and linter show no problems (`npm t`)
- [ ] tests are added/updated for bug fixes and new features
- [x] code is properly formatted (`npm run fmt`)
- [ ] description of changes is added in CHANGELOG.md
- [ ] update .d.ts typings

PR #1490, opened 2021-02-21
Procedure execution timeout

vara855:
- [x] tests and linter show no problems (`npm t`)
- [x] tests are added/updated for bug fixes and new features
- [x] code is properly formatted (`npm run fmt`)

Closes: #1481 

@tshemsedinov help me with next questions: 
* do i need to make timeoutFn as private class method, or move it somewhere to utils? 
* do i need to make a custom error class for timed out case? 

I'll make a changes from your suggestions. Have a nice day, this tech stack is awesome 👏

PR #1949, opened 2023-12-19
Reimplement global timeouts.request setting usage during Procedure invocation

KLarpen:
This PR contains latest rebased version of commit from  https://github.com/metarhia/impress/pull/1948 that was closed in intermediary state.

Updated and added test cases to control the behavior:
- if application env has no timeouts.request in the config Procedure fallback to 0 timeout
- config.server.timeouts.request applied to procedure script execution
- local timeout setting of the script overrides global timeouts.request  in both directions.

Closes: https://github.com/metarhia/impress/issues/1947 
Note: the issue already closed by mistake
Refs: https://github.com/metarhia/impress/pull/1948

- [x] tests are added/updated for bug fixes and new features
- [x] description of changes is added in CHANGELOG.md
- [ ] update .d.ts typings

### Situation investigation

The use of config.server.timeouts.request setting originated from commit  "Read timeouts from config" https://github.com/metarhia/metacom/pull/166 . That time it was used in metacom `lib/server.js` as a general timeout of any request processing. 

Setting was used in metacom `Server` until commit "Remove duplicated rpc call timeout" https://github.com/metarhia/metacom/commit/6062718a452e8f76ddcb7fb2e8775cbb79a8e728 where request level timer was completely removed from metacom `Server` in favor of newly implemented local timeout of a "Procedure execution" https://github.com/metarhia/impress/pull/1490 .

However `Procedure` since that time till now had used only local script level `timeout` setting, so the config.server.timeouts.request value wasn't applied to anything.

Based on that previously made decisions I decided that global timeouts.request setting must be used as default value for `Procedure` execution in case of absent local `timeout` setting. On the other side, existing local `timeout` value of a script, in my opinion should override the global: making timeout shorter or even longer than global.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased][unreleased]
 
 - Fixed API endpoints local queue settings applying
+- Reimplement global timeouts.request usage during a Procedure invocation
 
 ## [3.0.13][] - 2023-10-22
 
--- a/lib/procedure.js
+++ b/lib/procedure.js
@@ -31,7 +31,8 @@ class Procedure {
     this.description = exp.description || '';
     this.access = exp.access || '';
     this.validate = exp.validate || null;
-    this.timeout = exp.timeout || 0;
+    const { timeouts } = application.config.server;
+    this.timeout = (exp.timeout ?? timeouts.request) || 0;
     this.serializer = exp.serialize || null;
     this.protocols = exp.protocols || null;
     this.deprecated = exp.deprecated || false;
--- a/test/api.js
+++ b/test/api.js
@@ -14,6 +14,7 @@ const application = {
   absolute(relative) {
     return path.join(this.path, relative);
   },
+  config: { server: { timeouts: {} } },
 };
 
 metatests.testAsync('lib/api load', async (test) => {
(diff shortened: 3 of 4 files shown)


KLarpen:
@tshemsedinov this PR contains latest rebased commit version (with resolved conflicts) which necessary for landing to say that https://github.com/metarhia/impress/issues/1947 issue is really closed.

KLarpen:
### QA

As I see `Procedure` class is used for normal API endpoints, webhooks and plugin based units (e.g. https://github.com/metarhia/Example/blob/master/application/api/geo.1.js ). Domain or lib methods triggered to execute in worker threads by `application.invoke` is not a Procedures so the timeouts.request limit will not affect them directly. However if `application.invoke` was called from API endpoint, the  timeouts.request can still abort request with timeout (even though worker thread will complete the long running execution later).  Am I right @tshemsedinov ?

So reimplementation of a global request timeout by this PR will imply limit on their default execution time.  How we may be sure, that no one currently relies on "existed" but not actually applied timeouts.request limit? Should we care about that?

KLarpen:
#### Results of my manual testing with Example project

API endpoints responses on exceeded global timeout:
- WS transport `{message: "Timeout of 5000ms reached", code: 408}`
- HTTP transport `{message: "Timeout of 5000ms reached", code: 408}` with status code "408 Request Timeout"
- webhook endpoint `{"message":"Internal Server Error","code":500}` with status code 500

@tshemsedinov Does that normal expected behavior of a webhooks?

tshemsedinov (approved):

PR #2010, opened 2026-04-05
Security: Arbitrary sandbox method execution via unvalidated invoke path

tuanaiseo:
## Problem

Worker message handling resolves and executes `data.method` directly using `namespaceByPath(sandbox, method)` without any allowlist or namespace restriction. If an attacker can influence scheduled task `run` values or invoke messages, they can call dangerous globals exposed in sandbox (for example `process.*` or `node.child_process.*`), resulting in remote code execution or denial of service.

**Severity**: `critical`
**File**: `lib/worker.js`

## Solution

Restrict callable targets to a strict allowlist (for example only `api.*` procedures), reject method paths outside approved prefixes, and enforce authorization checks before dispatch. Do not expose powerful globals (`process`, full `node` namespace) to callable contexts.

## Changes

- `lib/worker.js` (modified)
- `lib/application.js` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/lib/application.js
+++ b/lib/application.js
@@ -19,7 +19,7 @@ const UserApplication = class Application extends EventEmitter {};
 
 const { COMMON_CONTEXT } = metarhia.metavm;
 const ERRORS = { Error, DomainError };
-const NAMESPACES = { node, npm, metarhia, process };
+const NAMESPACES = { npm, metarhia };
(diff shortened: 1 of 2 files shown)


tshemsedinov:
ai slop

PR #2022, opened 2026-08-13
WIP: Support namespace prefixes

kan0neego:
- Remove direct assignment of original name; only camelCase key is used
- Normalize by replacing '/' and '_' with '-' before camelCase conversion
- Strip `node:` prefix for built-in modules
- Remove explicit aliases for childProcess, StringDecoder, perfHooks, asyncHooks, and timers/promises
- Add test to ensure no module keys contain forbidden characters (`node:`, `-`, `_`)

- [X] tests and linter show no problems (`npm t`)
- [X] tests are added/updated for bug fixes and new features
- [X] code is properly formatted (`npm run fix`)
- [ ] update .d.ts typings


tshemsedinov (changes requested):
The direction looks good.

A few changes are needed before merge:
- "name.splice(1)" should be "name.slice(1)"
- The new test doesn't actually check forbidden characters
- I would like to have both normalized and not normalized names
- I suggest making the normalization rule explicit and consistent:

```
node:fs/promises      -> node.fsPromises
node:child_process    -> node.childProcess
node:string_decoder   -> node.stringDecoder
node:stream/consumers -> node.streamConsumers
```

Then we can remove all special aliases, including "StringDecoder" and "fsp".

tshemsedinov:
Also we do not use semantic commit messages like feat: etc.

tshemsedinov:
I'll land this after fixes with commit nama: Support namespace prefixes

